PR #1108, opened 2019-03-04
Fix style of ContextMenu after applying decoration

saarmbruster:
User-defined stylesheets for context menus do not work anymore after applying a decoration using `Decorator.addDecoration(...)`.

The Decorator applies a `DecorationPane` as root element in the scene graph but does not apply the stylesheets from the old root. While this works fine for most nodes, it does not work on user defined context menus. The context menus seem to only recognize the stylesheets of the root element.

I was only able to reproduce this issue when the nodes of the UI were loaded from an FXML file. When the nodes were defined in Java code, it worked fine.

This fix reads the stylesheets from the old root and applies them to the new `DecorationPane`.

The attached zip file cotains a sample project that reproduces this issue. The red button has a context menu with green background. If the button "Enable validation" is clicked, a validator will be attached to the text field. However, the context menu of the red button then loses it's custom style.

 The example can be run using `mvn compile exec:java`.

[sample.zip](https://github.com/controlsfx/controlsfx/files/2925838/sample.zip)

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/decoration/Decorator.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/decoration/Decorator.java
@@ -195,7 +195,8 @@ public class Decorator {
                 if (_pane == null) {
                     currentlyInstallingScenes.add(scene);
                     _pane = new DecorationPane();
-                    Node oldRoot = scene.getRoot();
+                    Parent oldRoot = scene.getRoot();
+                    _pane.getStylesheets().addAll(oldRoot.getStylesheets());
                     ImplUtils.injectAsRootPane(scene, _pane, true);
                     _pane.setRoot(oldRoot);
                     currentlyInstallingScenes.remove(scene);


JonathanGiles:
This should probably go into the master branch, but @abhinayagarwal - can you please clarify?

abhinayagarwal:
Hi,

I haven't been able to reproduce this issue. Steps I followed:

1. Run application on Java 11 using `mvn compile exec:java.`
2. Right click on the Textfield to show the green ContextMenu.
3. Enable validation
4. Right click on the Textfield and it still shows the ContextMenu with green background.

Here is a snapshot of the application with the ContextMenu after enabling the validation on the TextField.

ControlsFX

saarmbruster:
Sorry my bad, maybe my description on how to reproduce the issue was a bit unclear. The context menu on the TextField does indeed still work after enabling the validation. But this context menu is the default one provided by the TextField. The red button also contains a context menu but this one is explicitly added to the button. After enabling the validation the background of this context is not green anymore.

Your steps were almost right, just open the context menu of the red button instead of the TextField to see the described issue.

Here is a snapshot showing the context menu before and after the validation.
before_after

abhinayagarwal:
Thank you for your patience. The fix works. Can you please update the PR (or raise a new PR) to fix the master branch instead of 9.0.0?

saarmbruster:
That's all fine. I'm glad that you had the time to review the fix. I was not able to change the source branch, so I created a new PR (https://github.com/controlsfx/controlsfx/pull/1143). Since we do not need this PR anymore, I have closed it.

PR #1110, opened 2019-03-05
Updated ValidationMessage class to support ok and info severity

SimonScholz:
Hi,

I've added two new severity types:
* ok
* info 

and updated the ValidationMessage class to make use of the new severity types.

I also adjusted the classes, which make use of the ValidationMessage and Severity classes.


controlsfx/src/main/java/org/controlsfx/tools/SVGLoader.java
```diff
@@ -42,9 +42,6 @@
 import javafx.stage.Stage;
 import javafx.util.Callback;
 
-import com.sun.javafx.webkit.Accessor;
```
abhinayagarwal: I don't think these changes belong to this PR. Can you please revert these?

controlsfx/src/main/java/org/controlsfx/validation/Severity.java
```diff
@@ -30,6 +30,8 @@
  * Defines severity of validation messages 
  */
 public enum Severity {
+	OK,
```
abhinayagarwal: Fix alignment

controlsfx/src/main/java/org/controlsfx/validation/decoration/GraphicValidationDecoration.java
```diff
@@ -44,84 +43,122 @@
...
- * <br> <br>
- * <img src="GraphicValidationDecorationWithTooltip.png" alt="Screenshot of GraphicValidationDecoration">  
+ * <img src="GraphicValidationDecorationWithTooltip.png" alt="Screenshot of
+ * GraphicValidationDecoration">
  * 
  */
 public class GraphicValidationDecoration extends AbstractValidationDecoration {
 
-    // TODO we shouldn't hardcode this - defer to CSS eventually
+	// TODO we shouldn't hardcode this - defer to CSS eventually
```
abhinayagarwal: Please make sure all files are indented with 4 spaces instead of Tabs

SimonScholz:
@abhinayagarwal Thanks a bunch for the review and sorry for providing a PR in such a bad condition.
I hope I've made the changes compliant now.
I've also signed the CLA.

SimonScholz:
Once this PR is merged I intend to contribute even more fixes/features for the validation.
I built a ControlValidationSupport class, which is also capable of aggregating validation messages and the severity. This is really useful in a wizard to show messages and for example bind the finish button to an aggregated validation state and so on.

controlsfx/src/main/java/org/controlsfx/validation/decoration/GraphicValidationDecoration.java
```diff
@@ -99,14 +111,36 @@ protected Node createDecorationNode(ValidationMessage message) {
         return label;
     }
 
+    protected Node getGraphicBySeverity(Severity severity) {
+        switch (severity) {
+        case ERROR:
+            return createErrorNode();
+        case WARNING:
+            return createWarningNode();
```
abhinayagarwal: `createErrorNode` and `createWarningNode` are deprecated. May be we should avoid using the same?

controlsfx/src/main/java/org/controlsfx/validation/ValidationMessage.java
```diff
@@ -84,6 +84,35 @@ public static ValidationMessage error( Control target, String text ) {
...
+    public static ValidationMessage ok( Control target , String text ) {
+    	return new SimpleValidationMessage(target, text, Severity.OK);
+    }
+
+    /**
+     * Factory method to create a simple ok message 
+     * @param target message target
+     * @return ok message
+     */
+    public static ValidationMessage ok( Control target ) {
```
abhinayagarwal: When would a user use this method? There is no message while the documentation says its supposed to show a message.

controlsfx/src/main/java/org/controlsfx/validation/ValidationMessage.java
```diff
@@ -84,6 +84,35 @@ public static ValidationMessage error( Control target, String text ) {
...
+    	return new SimpleValidationMessage(target, text, Severity.INFO);
+    }
+
+    /**
+     * Factory method to create a simple ok message 
+     * @param target message target
+     * @param text message text 
+     * @return ok message
+     */
+    public static ValidationMessage ok( Control target , String text ) {
```
abhinayagarwal: The new `INFO` severity is a great addition. Thanks for stepping up for this functionality.

However, I am not able to get around the idea of `OK` severity. Can you please describe as to when this can be useful?

SimonScholz: I've created a Class named `ControlValidationSupport,` which tracks the validation state of a control and this class has a messageProperty of type `ObjectProperty<ValidationMessage>` and an invalidProperty of type  `BooleanProperty`.

When I have multiple ControlValidationSupport instances for different controls I want to be able to aggregate these messages and show the message with highest severity to the user, for example in the title or heading of a wizard. 
Once all validations are fine I want to display the default or "ok" message and turn the invalid state to false. In my coding I bind the aggregated invalidProperty of _ControlValidationSupport_ instances to the finish button of a wizard.

Does that make sense to you?

controlsfx/src/main/java/org/controlsfx/validation/ValidationMessage.java
SimonScholz: See answer in the follow up question...

controlsfx/src/main/java/org/controlsfx/validation/decoration/GraphicValidationDecoration.java
SimonScholz: Well in my former commit I actually avoided it, but then thought that it is not backwards compatible any more, because I actually added the @deprecated annotation to these methods.
If I still call these methods in our implementation it won't break existing code.
I don't know which rules you have when to remove/delete @deprecated methods,  but until then I think we should still call the @deprecated methods until we decide to remove these methods.

Agree?

abhinayagarwal: Sounds good. I was just being precautious :)

controlsfx/src/main/java/org/controlsfx/validation/ValidationMessage.java
abhinayagarwal: The OK severity will only be used when all the other validations pass. This can either just remove the previous validation or add its own message. Is that right?

SimonScholz: Yes that's correct.
I would then consider the ValidationMessage more like a Status class, which represents the overall status of a validation.
In Eclipse RCP wie also have a similar IStatus, which also contains a Severity and a Message.
And it also has OK, INFO, WARNING and ERROR.

Let me know if you still have concerns about this PR.

abhinayagarwal:
I am going ahead and merge this PR. 

I have created #1163 as a follow up on samples.

Issue #1114, opened 2019-03-08
Old notifications shown are not aligned with new notifications when the window is moved


Ciruman:
How to reproduce:

1. Start HelloNotifications.java standalone application
2. Choose "Set Owner"
3. Change fade delay to maximum
4. Show a notification
5. Move the application window
6. Show another notification

Current status:
The new notification will be shown wrongly aligned.

Expected:
The new notifications are aligned with the new ones.

Ciruman:
#1116 will also fix the first time shown, currently wrong.

Ciruman:
More information: https://www.ui-development.com/blog/2019/3/22/controls-fx-notification-pane-fix

PR #1132, opened 2019-04-05
#1114 Notifications: Fix notifications position

Ciruman:
The position of the elements does not use the notification position
anymore. It uses the position where the notifications will be shown.

The following errors are fixed with this change:
- Fix old notifications shown not aligned with new notifications
when the window is moved or owner is set.
- Fix animation, it does not use the position of the notification
because if the animation is not finished, the position is not
correct. Previously, the position between the first and second
notification was not right if there were several notifications.

This is the fix for the problem in Notifications position.

Issue: #1114(#1114)

New features/fixes easily reproducible with HelloNotifications class:

New notifications are aligned with the old ones.
All notifications are correctly aligned.
the functionality was tested in GNU/Linux, Windows and Mac. In Windows, if the notification is shown with an owner, and the owner is moved, the following notifications will be correctly aligned.

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/Notifications.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/Notifications.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014, 2016, ControlsFX
+ * Copyright (c) 2014, 2019, ControlsFX
  * All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


Ciruman:
More information: https://www.ui-development.com/blog/2019/3/22/controls-fx-notification-pane-fix

Issue #1135, opened 2019-04-12
TableFilter causing "Invalid URL" error


danielpeintner:
I created a simple test project (see https://github.com/danielpeintner/Java11Test) to explore functionality of upcoming Java11 release.
Hence I used `'org.controlsfx:controlsfx:11.0.0-RC2'` dependency.

I run into one issue with `TableFilter`. The line

`private static final Image filterIcon = new Image("/impl/org/controlsfx/table/filter.png");`
see https://github.com/controlsfx/controlsfx/blob/9.0.0/controlsfx/src/main/java/org/controlsfx/control/table/FilterPanel.java#L72

caused the following error

```
Caused by: java.lang.IllegalArgumentException: Invalid URL: Invalid URL or resource not found
	at javafx.graphics/javafx.scene.image.Image.validateUrl(Image.java:1107)
	at javafx.graphics/javafx.scene.image.Image.<init>(Image.java:617)
	at org.controlsfx.controls/org.controlsfx.control.table.FilterPanel.<clinit>(FilterPanel.java:72)
```

danielpeintner:
I can propose a simple fix by replacing

```
    private static final Image filterIcon = new Image("/impl/org/controlsfx/table/filter.png");

    private static final Supplier<ImageView> filterImageView = () -> {
        ImageView imageView = new ImageView(filterIcon);
        imageView.setFitHeight(15);
        imageView.setPreserveRatio(true);
        return imageView;
    };
```

with

```
    private static final Supplier<javafx.scene.Node> filterImageView = () -> {
		org.controlsfx.glyphfont.Glyph gFilter = new org.controlsfx.glyphfont.FontAwesome().create(org.controlsfx.glyphfont.FontAwesome.Glyph.FILTER);
    	return gFilter;
    };
```

Shall I propose this fix to the `master` or the `9.0.0` branch?
It uses fontawesome icons which is a better fit anyway. Hope this is ok?

abhinayagarwal:
@thomasnield Do you have a preference for the icon?

danielpeintner:
FYI: I created a PR using FontAwesome icon (https://github.com/controlsfx/controlsfx/pull/1138)

IF this icon does not work please let me know...

abhinayagarwal:
The preferred fix is not correct. The code doesn't work because the resource is encapsulated by the module system. An easy fix to the problem would be to load the resource using the classloader:

```
[...]

PR #1138, opened 2019-04-15
fix: issue 1135

danielpeintner:
see https://github.com/controlsfx/controlsfx/issues/1135

Note: uses FontAwesome filter icon.


abhinayagarwal:
@danielpeintner Please check my reply on the issue. Would you like to update the PR or should I close it?

danielpeintner:
I will provide an updated PR soon ...

PR #1139, opened 2019-04-16
Make javafx.web and javafx.media as optional modules

Eng-Fouad:
I faced an issue in my project where I am not using javafx.web. I got the following error on application startup:

> java.lang.module.FindException: Module javafx.web not found, required by org.controlsfx.controls

Making this module as optional module solves the problem.


abhinayagarwal:
This is excellent. Thank you!

Issue #1140, opened 2019-04-19
NotificationPopupHandler.getScreenBounds throwing NPE on 11.0.0


nicolabeghin:
Hi everyone
I just updated to ControlsFX 11.0.0 (from the RC2) and I'm getting some random reports of the NPE as below. Any suggestion? Unfortunately I can't replicate.

```
java.lang.NullPointerException
	at org.controlsfx.control.Notifications$NotificationPopupHandler.getScreenBounds(Notifications.java:364)
	at org.controlsfx.control.Notifications$NotificationPopupHandler.show(Notifications.java:343)
	at org.controlsfx.control.Notifications.show(Notifications.java:305)
	at org.controlsfx.control.Notifications.showInformation(Notifications.java:280)

```

thanks
nicola

nicolabeghin:
At the moment I had to downgrade to the RC2

abhinayagarwal:
@Ciruman Can this be caused because of #1130 ?

Ciruman:
> @Ciruman Can this be caused because of #1130 ?

Yes, I think so. I will have a look.

Ciruman:
Firstly, I found the error. ```Utils.getWindow(null)``` can return null. I will fix it now.

nicolabeghin:
Thanks a lot!

PR #1149, opened 2019-04-30
#1140 Fix NPE in Notifications

Ciruman:
Utils.getWindow(null) can return null and therefore we need to check if
there is a Window or we use primary window.


SubZeroPL:
Shouldn't that also be included in branch 9.0.0? As this error is still occurring there.

Maxoudela:
I thought that modifications made on `master `were merged also in `9.0.0` branch so that should be the case..

PR #1155, opened 2019-05-17
Wrap content in exception dialog

tobiasdiez:
If you set a preferred width for the exception dialog, then the content message is simply cut-off if it is too long. 
image

With this fix, the content should wrap nicely.


falkoschumann:
Without preferred width the content label will not wrap, only expand the dialog.

abhinayagarwal:
@falkoschumann Please file a new bug if you think there something to be fixed.

Issue #1159, opened 2019-05-28
Features page linked from README.md is missing


jimmoores:
The separate features page linked to by the README.md file appears to be missing. I therefore can't see what features controls fx has!

abhinayagarwal:
Feel free to create a PR

jeacott1:
how would a controlsfx newbie create a features page? - the features page that is linked bounces back to to this github. nobody can see what controlsfx does.

jimmoores:
I managed to miss the initial reply. I find it hilarious that I showed interest in this project, couldn't see what features it had, and rather than just clicking away, reported it, and was told to do it myself. Look, I maintain my own open source projects, I realise that I'm not entitled to support, but if anybody cares about bringing new users to this project at all, someone should fix  this, and I don't have the knowledge to do it.

JonathanGiles:
The old features page is here: http://fxexperience.com/controlsfx/features/
If someone was kind enough to transfer this into GitHub, that would be great.

jeacott1:
@JonathanGiles it seems someone has just now made that link accessible.
very helpful thanks.

JonathanGiles:
Yes, me, for the purpose of having someone move it to GitHub - then I plan to take it back down. It is better in the centrally managed github repo rather than on my website.

jeacott1:
if its your content isnt it easiest for you to copy it to the repo here rather than someone scraping it from the web?

JonathanGiles:
I'm fine with it being scraped from the web. It's not more easy for me than anyone else.

jeacott1:
you have all the images and whatever already in a folder. somewhere. pulling it from the web never works so cleanly,

JonathanGiles:
They are all sitting on a web server I don't have file access to. I would need to pull them down from the web in the same way.

jimmoores:
Now I have access to the source material, I will submit a PR.

jimmoores:
#1173 has been submitted.

abhinayagarwal:
Added the page to ControlsFX wiki.

PR #1173, opened 2019-07-12
#1159 fixed missing features page and converted source to MarkDown

jimmoores:
Given that source was embedded in a blog, I removed surrounding HTML and discarded CSS and JS.  I then converted the HTML to Markdown and cleaned up the results by hand. Fixed the image links, removed unused images, etc.  One remaining broken link is to a page containing dialogs, but that doesn't appear to be accessible so I can't include it.


abhinayagarwal:
Thanks for taking the effort to do this.

However, this should be a part of the Wiki. Please give me sometime to copy this to the Wiki, post which I will close the PR.

jeacott1:
@abhinayagarwal I don't think there's any harm adding this to the source. it might help keep it updated.

abhinayagarwal:
We are trying to consolidate all the information into the Wiki. I have updated the [features page.](https://github.com/controlsfx/controlsfx/wiki/ControlsFX-Features)

Once again, thank you for this :)

PR #1174, opened 2019-07-20
Fix build with Gradle 5

### LINKED ISSUE #1172: doesnt build with gradle 5.5
the deprecated left shift used here
```
    // Used in Travis
    task getVersion  << {
        println "$project.version"
    }
```

breaks the build. comment out the left shift and it works ok.

martinfrancois:
Fix #1172

The left shift operator (`<<`) is deprecated since Gradle 3.2 and was removed from Gradle 5. With the left shift operator, controlsfx will not build using Gradle 5. I replaced it with the `doLast` block, which makes it build again with Gradle 5.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,8 +74,10 @@ subprojects {
     }
 
     // Used in Travis
-    task getVersion << {
-        println "$project.version"
+    task getVersion {
+        doLast {
+            println "$project.version"
+        }
     }
 
 }


martinfrancois:
Can somebody please re-run the failing build? I don't think the failure is related to my changes, as the other one is passing.

abhinayagarwal:
@martinfrancois The failed Travis pipeline shouldn't have run at all. I had missed to remove the same from the repository settings. The PR should pass [here](https://travis-ci.org/controlsfx/controlsfx/) and it does.

PR #1194, opened 2019-10-09
Introducing RichText feature for SpreadsheetView

Maxoudela:
The general idea is to have some `RichText` inside the `SpreadsheetView`. A `RichText` is a Text that can have some parts in bold, italic. Different font size, font family etc.

One option was to use the `TextFlow` provided by JavaFX but we lack an editor to edit it for end user. So we decided to go for the `HTMLEditor` for edition, that will provide some HTML text. This HTML will then be rendered inside the `SpreadsheetView` using `WebView` allowing the HTML to be well rendered. 

`RichText` cell can be made using the new `CellGraphicFactory` and providing some WebView to use inside the cell.

Since `WebView` can consume some memory, we do not set the `WebView` directly in the `SpreadsheetCell` but provide the `CellGraphicFactory`. `CellView` will then return the `WebView` when not used anymore (needs improvements as some may be lost apparently). 

I've decided not to enforce the use of `WebView` in my API. Therefore, the factory is called `CellGraphicFactory` and can be used to provide anything into the cells that would need to be given and returned. I'm aware that this is not optimal, we are reaching the limits of the `SpreadsheetCell` model and introducing the initial `TableCellFactory` to offer full control on the TableCells for the user would be a good thing to do.

I have implemented a bunch of unit tests on my project, I'll try to add them into ControlsFX but I'm making use of TestFX and the headless mode so it may be a bit tricky.


Maxoudela:
Should I squash these commits into one maybe?

abhinayagarwal:
Hi Sam!

I have never used the SpreadsheetView so I am a bit reluctant to review this PR. @jperedadnr or @JonathanGiles might be good reviewers for this. If they are busy, I can approve this on good faith :)

As for the squash, we could do it via Github once the PR is ready to be merged.

Maxoudela:
There is a lot of modifications, I'll try to modify the pull request in order to explain the modifications as clearly as I can.

controlsfx/src/main/java/org/controlsfx/control/spreadsheet/SpreadsheetView.java
```diff
@@ -407,7 +404,7 @@
     
     //Zoom for the SpreadsheetView.
     private DoubleProperty zoomFactor = new SimpleDoubleProperty(1);
-    private static final double MIN_ZOOM = 0.1;
+    private static final double MIN_ZOOM = 0.2;
```
Maxoudela: Zooming to 10% was really low, I've decided to push it a little bit more.

controlsfx/src/main/java/org/controlsfx/control/spreadsheet/SpreadsheetView.java
```diff
@@ -1093,21 +1111,23 @@ public final DoubleProperty zoomFactorProperty() {
...
     public void incrementZoom() {
-        double newZoom = getZoomFactor();
-        int prevValue = (int) ((newZoom - MIN_ZOOM) / STEP_ZOOM);
-        newZoom = (prevValue + 1) * STEP_ZOOM + MIN_ZOOM;
-        setZoomFactor(newZoom > MAX_ZOOM ? MAX_ZOOM : newZoom);
+        double newZoom = zoomFactor.getValue() + STEP_ZOOM;
+        newZoom *= 10;
+        newZoom = Math.floor((float) newZoom);
+        newZoom /= 10;
+        zoomFactor.setValue(newZoom > MAX_ZOOM ? MAX_ZOOM : newZoom);
```
Maxoudela: Previous implementation was wrong. Simply fixing it.

controlsfx/src/main/java/org/controlsfx/control/spreadsheet/SpreadsheetView.java
```diff
@@ -1906,12 +1926,7 @@ public void copyClipboard() {
                  */
                 for (int row = 0; row < getRowSpan(cell, p.getRow()); ++row) {
                     for (int col = 0; col < getColumnSpan(cell); ++col) {
-                        try {
-                            new ObjectOutputStream(new ByteArrayOutputStream()).writeObject(cell.getItem());
-                            list.add(new ClipboardCell(p.getRow() + row, p.getColumn() + col, cell.getItem() == null ? null : cell.getItem()));
-                        } catch (IOException exception) {
-                            list.add(new ClipboardCell(p.getRow() + row, p.getColumn() + col, cell.getItem() == null ? null : cell.getItem().toString()));
-                        }
+                        list.add(new ClipboardCell(p.getRow() + row, p.getColumn() + col, cell));
```
Maxoudela: The exception part is now done into the ClipBoardCell

controlsfx/src/main/java/org/controlsfx/control/spreadsheet/SpreadsheetCell.java
```diff
@@ -138,7 +138,29 @@
...
+     * something particular and should be display by using a Node provided by
+     * {@link CellGraphicFactory} object in the CellView.
+     *
+     * If you only seek to place a simple {@code Node} for this cell, simply use {@link #setGraphic(javafx.scene.Node)
+     * }.
+     *
+     * @param isCellGraphic if {@code true}, a Node will be used to display
+     * something particular for the cell
+     */
+    public void setCellGraphic(boolean isCellGraphic);
```
Maxoudela: The API is a bit confusing, especially because `setCellGraphic` is existing. I tried to be explicit in the JavaDoc to differentiate both. `setGraphic` is simply a shortcut to the official JavaFX method. `setCellGraphic` is a new method only accepting a boolean, and the `CellView` once actually showed will request the graphic to show, and return the graphic once hidden.

controlsfx/src/main/java/impl/org/controlsfx/spreadsheet/GridRowSkin.java
```diff
@@ -230,41 +230,38 @@ protected void layoutChildren(double x, final double y, final double w, final do
...
-                 * condition is to determine that we are dealing with the very
-                 * first cell of a columnSpan. If we have the hidden cells, we
-                 * must not increase the fixedColumnWidth.
+                 * Here we verify if our cell must be shifted. We must increase
+                 * all cells that are not column_span_invisible because their
+                 * whole width will be computed by the original cell that is
+                 * row_visible.
                  */
-                if (hbarValue + fixedColumnWidth > x &&  spreadsheetCell.getColumn() == indexColumn) {
+                if (spanType != SpreadsheetView.SpanType.COLUMN_SPAN_INVISIBLE && hbarValue + fixedColumnWidth > x) {
```
Maxoudela: Another bug identified where cells in row_span invisible were hidden.

controlsfx-samples/src/main/java/org/controlsfx/samples/spreadsheet/BrowserImpl.java
```diff
@@ -0,0 +1,112 @@
...
+    private final Map<WebView, String> loadedUrl = new HashMap<>();
+
+    @Override
+    public Node getNode(SpreadsheetCell cell) {
+        //Limit the number of browser displayed.
+        if (browserCounter.get() < 256) {
+            if (browserList.isEmpty()) {
+                browserCounter.incrementAndGet();
+                WebView webView = new WebView();
+                webView.setMouseTransparent(true);
```
Maxoudela: The WebView must be mouseTransparent in order to let the cell be selected. Otherwise the SpreadsheetView could become unusable.

controlsfx-samples/src/main/java/org/controlsfx/samples/spreadsheet/BrowserImpl.java
```diff
@@ -0,0 +1,112 @@
...
+                webView.setContextMenuEnabled(false);
+                webView.setPrefHeight(48);
```
Maxoudela: The WebView cannot autofit itself based on the content. This is very annoying so in this example, I've set it to 48 (2 row) but a specific implementation should manage the webViewheight in another manner.

controlsfx-samples/src/main/java/org/controlsfx/samples/spreadsheet/BrowserImpl.java
```diff
@@ -0,0 +1,112 @@
...
+                //https://stackoverflow.com/questions/11206942/how-to-hide-scrollbars-in-the-javafx-webview?lq=1
+                webView.getChildrenUnmodifiable().addListener(new InvalidationListener() {
```
Maxoudela: Since the WebView is not responding to mouseEvent, we must hide the scrollBars that could be shown.

Issue #1163, opened 2019-06-01
Add samples for OK and INFO severity


abhinayagarwal:
Add samples for the changes done in PR #1110.

This can be done by updating HelloValidation class, if required.

Mason-Chou:
My PR only covers the bare minimum to work onto this issue. We might need to add more factory methods in ``Validator`` class for ease of use on the developer. From my experience, it's hard to add a "negated" condition as a validator. Example would be, I want an Error Severity on a TextField when it has no value but I want a INFO Severity when the user input the value **Test**.

PR #1196, opened 2019-10-22
added support for the INFO Severity and refactored the messages storage

Mason-Chou:
This is needed for Issue #1163. Currently the ValidationSupport model does not recognize messages that have the INFO Severity. This merge will add in that support.

Also refactor how messages are stored. Use one ArrayList instead of creating an Arraylist for each Severity type.

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/validation/ValidationResult.java
+++ b/controlsfx/src/main/java/org/controlsfx/validation/ValidationResult.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014, ControlsFX
+ * Copyright (c) 2014, 2019, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/controlsfx/src/main/java/org/controlsfx/validation/ValidationResult.java
+++ b/controlsfx/src/main/java/org/controlsfx/validation/ValidationResult.java
@@ -50,7 +50,8 @@ public class ValidationResult {
     public ValidationResult() {}
 
     /**
-     * Factory method to create validation result out of one message
+     * Factory method to create validation result out of one message.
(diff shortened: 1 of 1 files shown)


Maxoudela:
Thanks for the PR.

I do not understand where the info message are stored? It seems that the `getMessages(null)` is only returning `invalidMessages` so any INFO message added will not be returned?

Mason-Chou:
Before the changes, we have a multiple ArrayLists storing the **WARN** and **ERROR** ValidationMessage separately. I've consolidated this into one ArrayList and named it ```invalidMessage``` as I deemed it wasn't scalable.

I've changed the method ```add(ValidationMessage message)``` to add all ValdationMessage that does not have a ***null*** severity nor a ***OK*** severity.

Maybe the variable ```invalidMessages``` is confusing as I deemed any ValidationMessage without an **OK** severity to be invalid. If you have a better variable name, I would appreciate that suggestion.

Maxoudela:
Hum yes thanks, I did not understand at first look, my bad. The Info message are indeed recorded into the `invalidMessages`. The variable name seems ok for me as OK messages are not recorded in it.

controlsfx/src/main/java/org/controlsfx/validation/ValidationResult.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014, ControlsFX
+ * Copyright (c) 2014, 2019, ControlsFX
```
Maxoudela: Can you remove the last comma? As explained [here](https://github.com/controlsfx/controlsfx/wiki/Contributing-to-ControlsFX), the general format is :

> If you are editing an existing source file, the format should be , . For example, if the file was created in 2012, and it is currently 2014, it should say '2012, 2014'.

The rest seems good 👍

PR #1200, opened 2019-10-28
fix for bitbucket #748

FlorianKirmaier:
fixing https://bitbucket.org/controlsfx/controlsfx/issues/748/memory-leak-in-actionutils-permanent

Issue #748, opened 2015-07-09
PopOver content can be cut off after detaching it


JonathanGiles:
**[Original report](https://bitbucket.org/controlsfx/controlsfx/issue/557) by Dirk Lemmermann (Bitbucket: [dlemmermann](https://bitbucket.org/dlemmermann), GitHub: [dlemmermann](https://github.com/dlemmermann)).**

----------------------------------------

After detaching a popover it can happen that its content gets cut off. We are missing a call to popOver.sizeToScene().

PR #1202, opened 2019-10-29
fixing bitbucket #748 https://bitbucket.org/controlsfx/controlsfx/iss…

FlorianKirmaier:
original pull request: https://github.com/controlsfx/controlsfx/pull/1200
the original pull request was a bit buggy, so I fixed it and added unit tests

The new change adds unit-tests for memory leaks. 
It also adds a small Utilityclass for detecting memory leaks. Actually, I plan to make a library for unittesting for memoryleaks. But for this change, this utility class should be enough.

Calling `configureButton` twice on the same button, without unconfiguring the old Action is now forbidden and causes an Exception. This change was not, because the previous behaviour let to memory leaks and unpredictable behavior of the application.


FlorianKirmaier:
The last version changed every line in ActionUtils because of different linespaces.
I've fixed it and forced pushed the fixed version

controlsfx/src/test/java/org/controlsfx/control/action/MemoryLeakUtils.java
```diff
@@ -0,0 +1,83 @@
+package org.controlsfx.control.action;
+
+import com.sun.management.HotSpotDiagnosticMXBean;
```
Maxoudela: I'm unsure about this import. Do we want to import these into ControlsFX? Will it cause issues when merged in JDK 11 branch?

The whole class is puzzling me. I did not know it was possible to test leak like that. Is it 100% sure? I thought the `System.gc();` output was unpredictable. I'm afraid this test adds 5 seconds to the build just to test for the listener. I'll let others review the changes.

FlorianKirmaier: I'm using this technique in all of my projects.
Without it, it's basically guessing, which parts of the application are leaking.
It's by spec unsafe, but this is manageable and becomes a mere theoretical discussion.

In my experience, without these tests, JavaFX-code usually gets full with memory leaks, just like this one. Without these tests, memory leaks usually only get noticed in production, which is a not very systematic way to work on software.

The test should only add about 0.1 seconds on average. (there was a small bug, which I will fix tomorrow)



Actually, I've just started to make a library out of it, because I needed it for another. And I finally had to stop to copy-paste this logic from project to project.

Currently, I'm working on publishing this library on MavenCentral.

It probably would sense, when I would use my library in ControlsFX and add a lot of unit-tests to the library itself, to make sure this approach is working properly.

I can also provide a version of the PR without unit-tests.

Maxoudela: I'm sorry, I have lots of work. I'll take a deep look asap (maybe next week). I think it's better to the PR with the unit test, I just want to be sure of what we have. But if we encounter some issue in JFX13 branch or 9.0.0 branch, then we will maybe remove it for this specific branch

FlorianKirmaier:
As mentioned, i moved the unittest class to an library which can now be used by different projects.
The library is named [JMemoryBuddy](https://github.com/Sandec/JMemoryBuddy)
This should simplify the pullrequest.
The library also contains some unit-tests to verify this kind of testing is working fine.
It's also tested against [openjdk13 with travis](https://github.com/Sandec/JMemoryBuddy/blob/master/.travis.yml)

controlsfx/src/test/java/org/controlsfx/control/action/TestActionUtils.java
```diff
@@ -0,0 +1,51 @@
+package org.controlsfx.control.action;
```
Maxoudela: Maybe add the ControlsFX copyright?

Maxoudela:
Sorry for the delay.
Apart from the copyright header in the test file, the whole thing seems good. 

I don't know the policy on foreign imports but I think it would be shame to remove tests because we don't want to import external libraries. This could open the door to importing TestFX library that is very nice for testing JavaFX application.

@abhinayagarwal  What is your opinion on that?

FlorianKirmaier:
Great, I've just added the copyright header and I've updated JMemoryBuddy/improved the naming of the methods.

Maxoudela:
Thanks for the PR, have you signed the [CLA](https://docs.google.com/forms/d/e/1FAIpQLSfGmJDMdl9mPBiZLwhUXPNRKcSSJDawBHe_ED2ReSJGUt-puQ/viewform) as mentioned in the [Contributing doc](https://github.com/controlsfx/controlsfx/wiki/Contributing-to-ControlsFX)?

FlorianKirmaier:
I've just signed it!

Maxoudela:
Thank you. On my side, this seems good but I'll other reviewers (@abhinayagarwal ?) give their opinion on the third party for test

abhinayagarwal:
Since this PR adds a "test" scope dependency and has minimum impact on the project, I am OK with this.

@FlorianKirmaier Can you please the conflict and update the header on `ActionUtils.java` ?

abhinayagarwal:
@FlorianKirmaier Does you library work with JDK 11? Just wondering if this addition will still be useful on `9.0.0` branch?

FlorianKirmaier:
@abhinayagarwal 
Yes, [JMemoryBuddy](https://github.com/Sandec/JMemoryBuddy) is tested against 8, 11 and 13.
Currently, no setup is known, which doesn't work with the library.

abhinayagarwal:
@FlorianKirmaier That's great. Please resolve the conflict so that we can merge this PR.

PR #1208, opened 2019-11-05
PR for Issue #1163

Mason-Chou:
- Added info severity to the HelloValidation sample in the ComboBox.
- Clean up code format and also removed excess comments
- Changed color of fx-effect to skyblue.
    -  I was getting a css error from the previous rbga value using **java8u211** on **Windows 10** 
    - [This JavaFX Bug has been confirmed to not be addressed in Java 8](https://bugs.openjdk.java.net/browse/JDK-8182152)

I didn't do anything for OK Severity. Not sure what the goal for this severity is for. I can add it in once I understand the intent of this. (Maybe use it to check when a user enter in a valid value?)


Mason-Chou:
Referencing Issue #1163

Maxoudela:
Thanks for the PR! I won't be able to help you for the OK severity as I don't use it :/

PR #1232, opened 2020-01-22
Bug Fixes for Issue #1213

Teletha:
When Notifications.show () is called when no Window has the focus, NPE is thrown. This is different from the behavior of the master branch, so I would like to fix it.

PR #1236, opened 2020-02-13
Bug Fixes for Issue #1213

### LINKED ISSUE #1213: NPE from Notifications.show() if app has no window open
If the app does not currently have a window open, this exception is thrown:

> Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
> 	at org.controlsfx.control.Notifications$NotificationPopupHandler.getScreenBounds(Notifications.java:364)
> 	at org.controlsfx.control.Notifications$NotificationPopupHandler.show(Notifications.java:343)
> 	at org.controlsfx.control.Notifications.show(Notifications.java:305)

If the app does have a window open, the problem does not occur.

In my case, the app is running in the background with only the SystemTray icon visible (AWT).

Teletha:
When Notifications.show () is called when no Window has the focus, NPE is thrown. This is different from the behavior of the master branch, so I would like to fix it.

Fixes #1213

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/tools/Utils.java
+++ b/controlsfx/src/main/java/org/controlsfx/tools/Utils.java
@@ -26,7 +26,7 @@
  */
 package org.controlsfx.tools;
 
-import java.util.List;
+import java.util.Iterator;
 
 import javafx.scene.Node;
 import javafx.stage.PopupWindow;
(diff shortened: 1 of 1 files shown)


abhinayagarwal:
Closing in favour of #1232

abhinayagarwal:
Could you elaborate on how this PR fixes #1213 ?

Maxoudela:
This fix is indeed reproducing the code of the master branch : https://github.com/controlsfx/controlsfx/blob/master/controlsfx/src/main/java/org/controlsfx/tools/Utils.java

But in case no stage is shown, `null` would still be returned and the NPE would still occurs if I'm not mistaken ?

controlsfx/src/main/java/org/controlsfx/tools/Utils.java
```diff
@@ -44,14 +44,16 @@
...
+            Iterator<Window> windows = Window.getWindows().iterator();
+            while (windows.hasNext()) {
+                window = windows.next();
                 if (window.isFocused() && !(window instanceof PopupWindow)) {
-                    return window;
+                    break;
                 }
             }
-            return null;
+            return window;
```
Maxoudela, line 56: In case no stage is shown, null would still be returned and the NPE would still occurs if I'm not mistaken ?

abhinayagarwal:
Closing this PR as the developer hasn't replied in a long time.

Issue #1119, opened 2019-03-21
Java11 OSGi Manifest imports cause errors in e(fx)clipse


SimonScholz:
Hi,

I am working on an e(fx)clipse rcp project and want to make use of ControlsFx, but the following line causes errors, since these imports are not really available and provided differently in the classpath by e(fx)clipse.

https://github.com/controlsfx/controlsfx/blob/272f5541dc35bf183acbc544016fe07099989ec2/controlsfx/build.gradle#L111

I think you should simply import nothing by omitting Line 111 from above and the problems should go away.

Maxoudela:
Related to #1191

Issue #1191, opened 2019-10-01
Clean up ControlsFX project


abhinayagarwal:
I would like to propose a clean up for ControlsFX project. There are many IDE specific and out-dated files which are no longer required.

Here is a list of files:

* [.project](https://github.com/controlsfx/controlsfx/blob/master/.project)
* [.classpath](https://github.com/controlsfx/controlsfx/blob/master/.classpath)
* [javafx.plugin](https://github.com/controlsfx/controlsfx/blob/master/javafx.plugin)
* [controlsfx-samples/eclipse/ControlsFX Samples.launch](https://github.com/controlsfx/controlsfx/blob/master/controlsfx-samples/eclipse/ControlsFX%20Samples.launch)
* [controlsfx-samples/.classpath](https://github.com/controlsfx/controlsfx/blob/master/controlsfx-samples/.classpath)
* [controlsfx-samples/.project](https://github.com/controlsfx/controlsfx/blob/master/controlsfx-samples/.project)
* [controlsfx/.classpath](https://github.com/controlsfx/controlsfx/blob/master/controlsfx/.classpath)
* [controlsfx/.project](https://github.com/controlsfx/controlsfx/blob/master/controlsfx/.project)
* [fxsampler/.classpath](https://github.com/controlsfx/controlsfx/blob/master/fxsampler/.classpath)
* [fxsampler/.project](https://github.com/controlsfx/controlsfx/blob/master/fxsampler/.project)
* [fxsampler/.hgignore](https://github.com/controlsfx/controlsfx/blob/master/fxsampler/.hgignore)

Also, there are some plugins which we use inside the [`build.gradle`](https://github.com/controlsfx/controlsfx/blob/master/build.gradle) file which can be removed:

* apply plugin: 'idea'
* apply plugin: 'eclipse'
* apply plugin: 'osgi' - Will need to remove 'instruction' in manifest { } from `controlsfx/build.gradle`. My experiene with OSGI is minimum. Therefore, I am not sure if removing it will have any adverse effect on the project.

eugener:
While I fully support removal of the files listed  (also we have to make sure they are all added to `.gitignore`), I'm not so sure the plugins need removal. They actually help IDEs to work correctly with the project. So I will keep them, unless there are issues with JPMS.

abhinayagarwal:
[...]

PR #1239, opened 2020-03-09
Add Export-Package for OSGi in MF

### LINKED ISSUE #1237: Missing Import-Package information in MF file.
In the 5f887f1 submission, I found that the jar is missing the description of Import-Package in MF. 

But in the manifest block, there is still this description: `the manifest of the default jar is of type OsgiManifest`. The submission did not say why it was removed.

I am using the controlsfx bundle and this information is important to me. If Import-Package must be removed, I can only export this information myself, I don't think it is necessary.

These changes affected 8.40.16.

sdlzhd:
Fixes #1237
In Gradle 5.0, the OSGi Plugin was deprecated. We need to modify the build scripts.

The details: [https://github.com/bndtools/bnd/tree/master/biz.aQute.bnd.gradle](https://github.com/bndtools/bnd/tree/master/biz.aQute.bnd.gradle)

### DIFF
--- a/controlsfx/build.gradle
+++ b/controlsfx/build.gradle
@@ -1,3 +1,7 @@
+plugins {
+    id 'biz.aQute.bnd.builder'
+}
+
 dependencies {
     try {
         jdk files(jfxrtJar)
@@ -12,7 +16,7 @@ ext {
 
 configurations {
     jdk
-    
+
     if (project.hasProperty("transifex.api")) {
         transifex_api = project.getProperty("transifex.api")
     }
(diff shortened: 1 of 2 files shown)


controlsfx/build.gradle
```diff
@@ -1,3 +1,7 @@
+plugins {
+    id 'biz.aQute.bnd.builder'
```
Maxoudela: This is https://github.com/bndtools/bnd right?

Apache License 2 so should be good for us.

Maxoudela:
I do not know anything on Gradle and OSGi so I'll let my peers judge this PR ;)

sdlzhd:
You're right. I don't know Gradle either, too. I just was able to package it right and it was recognized in OSGi.
This is great if someone is familiar with the Gradle plug-in

controlsfx/build.gradle
sdlzhd: Yeah!

abhinayagarwal:
I am in a similar spot as @Maxoudela here. No idea on OSGI. I will ask a review from @eugener on this. If he doesn't have time and can't review it, we can simply merge this PR as it seems to be a low risk PR.

abhinayagarwal:
I removed the OSGI plugin as a part of #1191. BTW, does this plugin also fix #1119 ?

eugener:
I don't have enough knowledge on OSGi to fully review, but if the plugin does the job - it is a low risk. Related issues can be reported/fixed by interested community members.

sdlzhd:
@abhinayagarwal It is due to #1191  that plugins are not available in OSGi. 

And starting from jdk11, javafx will be a separate module, but in jdk8, packages  prefixed with javafx do not need to be provided in the form of **import-package**. So in this PR, I checked the jdk environment.

PR #1242, opened 2020-03-22
Allow user to overwrite the rating icon

tobiasdiez:
As of now it not possible to easily customize the look of the rating control apart from changing the look via css. Other customization like using font-based glyphs for the star are not possible.

This PR makes the creation of the star node protected so that one customize it by overriding.

### DIFF
--- a/controlsfx/src/main/java/impl/org/controlsfx/skin/RatingSkin.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/skin/RatingSkin.java
@@ -259,14 +259,12 @@ public class RatingSkin extends BehaviorSkinBase<Rating, RatingBehavior> {
     	
     }
     
-//    private double getSpacing() {
-//        return (backgroundContainer instanceof HBox) ?
-//                ((HBox)backgroundContainer).getSpacing() :
-//                ((VBox)backgroundContainer).getSpacing();
-//    }
+    protected Node createButtonNode() {
+        return new Region();
+    }
     
     private Node createButton() {
-        Region btn = new Region();
+        Node btn = createButtonNode();
         btn.getStyleClass().add("button"); //$NON-NLS-1$
         
         btn.setOnMouseMoved(mouseMoveHandler);
--- a/controlsfx/src/main/java/impl/org/controlsfx/skin/RatingSkin.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/skin/RatingSkin.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013, 2015 ControlsFX
+ * Copyright (c) 2013, 2015, 2020 ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


controlsfx/src/main/java/impl/org/controlsfx/skin/RatingSkin.java
```diff
@@ -259,14 +259,12 @@ private void updateClip() {
     	
     }
     
-//    private double getSpacing() {
-//        return (backgroundContainer instanceof HBox) ?
-//                ((HBox)backgroundContainer).getSpacing() :
-//                ((VBox)backgroundContainer).getSpacing();
-//    }
+    protected Node createButtonNode() {
```
abhinayagarwal: Please add Javadoc on this method.

abhinayagarwal:
Hi @tobiasdiez ,

Thanks for the PR, have you signed the [CLA](https://docs.google.com/forms/d/e/1FAIpQLSfGmJDMdl9mPBiZLwhUXPNRKcSSJDawBHe_ED2ReSJGUt-puQ/viewform) as mentioned in the [Contributing doc](https://github.com/controlsfx/controlsfx/wiki/Contributing-to-ControlsFX)?

controlsfx/src/main/java/impl/org/controlsfx/skin/RatingSkin.java
abhinayagarwal: Please update the license year in the source file.

tobiasdiez:
Thanks for the review. Feedback is incorporated and CLA signed.

controlsfx/src/main/java/impl/org/controlsfx/skin/RatingSkin.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013, 2015 ControlsFX
+ * Copyright (c) 2013, 2015, 2020 ControlsFX
```
abhinayagarwal, line 2: Please update this to "Copyright (c) 2013,, 2020, ControlsFX"

controlsfx/src/main/java/impl/org/controlsfx/skin/RatingSkin.java
```diff
@@ -259,14 +259,17 @@ private void updateClip() {
...
-//        return (backgroundContainer instanceof HBox) ?
-//                ((HBox)backgroundContainer).getSpacing() :
-//                ((VBox)backgroundContainer).getSpacing();
-//    }
+    /**
+     * Creates the node for one star.
+     * By default, an empty region is returned.
+     * This method can be overwritten to customize the styling of the star, e.g. to use a font-based icon.
+     */
+    protected Node createButtonNode() {
```
abhinayagarwal, line 267: A more generic name for this method is "display node" instead of "button node" i.e. `createDisplayNode`. Although we use the style-class "button", in theory it is not a button.

tobiasdiez: I agree with you. However, it reacts as some form of button (has onclick action) and is called a button everywhere else in code.

abhinayagarwal: I would like @Maxoudela to pitch in his view.

Maxoudela: Sorry for the delay.
The button naming is maybe a bit wrong but it's indeed called "buttons" everywhere in the file so I would say to keep it that way for consistency.

PR #1251, opened 2020-04-11
Remove SVGLoader and update build files

### LINKED ISSUE #1250: Remove SVGLoader
[SVGLoader](https://github.com/controlsfx/controlsfx/blob/master/controlsfx/src/main/java/org/controlsfx/tools/SVGLoader.java) is the only class in ControlsFX which has a dependency on `javafx.web` module. Removing this would remove ControlsFX dependency on this module.

It also tries to access [WebPage](https://github.com/controlsfx/controlsfx/blob/master/controlsfx/src/main/java/org/controlsfx/tools/SVGLoader.java#L112) which is an internal API due to which we have to fall back on reflection.

By removing this tool from ControlsFX, we can ensure that the library is trimmed down and only uses what is most essential without depending on any internal API.

abhinayagarwal:
Fixes #1250


Maxoudela:
The new feature of HTML cell in SpreadsheetView is using Webview, but only in samples I recall. There is one here (https://github.com/controlsfx/controlsfx/blob/master/controlsfx/src/main/java/org/controlsfx/control/spreadsheet/CellGraphicFactory.java) but it's javadoc so it can be removed. Anyway these are not present in 9.0.0 branch but in jfx-13, we should just be careful if we ever merge some commits from 9.0.0 to jfx-13.

abhinayagarwal:
That's just Javadoc, so I won't be worried about it. We will have to merge 9.0.0 into jfx-13 before we make a release for 13.

PR #1253, opened 2020-04-13
Removing JavaFXThreadingRule and using TestFX on tests instead.

### LINKED ISSUE #1252: Remove javafx.swing module
`javafx.swing` module is used by [JavaFXThreadingRule](https://github.com/controlsfx/controlsfx/blob/9.0.0/controlsfx/src/test/java/org/controlsfx/control/JavaFXThreadingRule.java). This class in turn is used by [CheckTreeViewTest](https://github.com/controlsfx/controlsfx/blob/9.0.0/controlsfx/src/test/java/org/controlsfx/control/CheckTreeViewTest.java).

IMO, it would be a good idea to get rid of both these classes and remove the dependency of ControlsFX on `javafx.swing`.

This will also provide us a **clean slate to introduce testing** for ControlsFX on 9.0.0 branch.

Maxoudela:
Fixes #1252


Maxoudela:
Allright, this is a first step. I want to go one step further by running tests in headless mode. For example using this :
```
static {
        System.setProperty("testfx.robot", "glass");
        System.setProperty("testfx.headless", "true");
        System.setProperty("prism.order", "sw");
        System.setProperty("prism.text", "t2k");
        System.setProperty("headless.geometry", "1600x1200-32");
    }
```

But I have this exception:
`java.lang.IllegalAccessError: superclass access check failed: class com.sun.glass.ui.monocle.MonoclePlatformFactory (in module org.testfx.monocle) cannot access class com.sun.glass.ui.PlatformFactory (in module javafx.graphics) because module javafx.graphics does not export com.sun.glass.ui to module org.testfx.monocle`

Seems related to https://github.com/TestFX/TestFX/issues/553 but I need to investigate further in order to resolve that.

abhinayagarwal:
@Maxoudela Thanks for the PR. I will do a follow-up PR and remove the `javafx.swing` module. 

I will also try and check on my end if I can enable headless mode by adding some JVM args to the test task.

PR #1273, opened 2020-05-19
9.0.0

Maxoudela:
Well, I first pushed directly to 9.0.0 (my bad) and I think I managed to revert that mistake.
Last time you merge master to 9.0.0 was 12 April 2020 so this should explain why there are so few commits right?

If this is good, I'll then try to merge 9.0.0 to jfx13 branch as we discussed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ dist: trusty
 install: true
 
 # Assemble jars
-script: ./gradlew clean assemble -Ptransifex.api=$TRANSIFEX_API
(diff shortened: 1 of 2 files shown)


abhinayagarwal:
LGTM!

I can see some people doing multiple commits in a PR which also gets in ControlsFX commit history. I like the policy in OpenJFX repository i.e. squash and merge every PR from, so that we create one commit per PR. Let's try to do that from now on :)

Having said that, for all merge from master to 9.0.0, we should always create a merge commit.

PR #1280, opened 2020-05-29
Merge 9.0.0 to jfx-13 for catch-up

Maxoudela:
This merge is a bit heavy, I may have done some mistakes in the `build.gradle` file since I resolved the conflict, but unsure of it.


abhinayagarwal:
Hi @Maxoudela ,

I have fixed and pushed a commit to the PR. This is now updated with all the changes in 9.0.0 branch. I just have one question, why are we still dependent on JavaFX 12? I thought you wanted to move to 13 as the latter had some additional API changes related to Spreadsheet and TableView.

Maxoudela:
Thanks a lot!

I don't know why I pushed the 12 version, maybe because the thing wasn't building and I tried some things. We should definitely use the 13 version, and maybe even 14 should work

abhinayagarwal:
Hi @Maxoudela ,

This PR is no longer required.

jfx-13 is now updated with all changes from 9.0.0

PR #1292, opened 2020-07-16
#1283 Add revalidate method

### LINKED ISSUE #1283: Provide method to force revalidation
It would be nice to have a method on `ValidationSupport` to force re-validation. Currently, the only way that validation is triggered is when the value changes. There are cases where the validity of a control can change even if the value of that control does not.

shollander:
Fixes #1283 for master (java 8)

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/validation/ValidationSupport.java
+++ b/controlsfx/src/main/java/org/controlsfx/validation/ValidationSupport.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014, 2015, ControlsFX
+ * Copyright (c) 2014, 2015, 2020, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


controlsfx/src/main/java/org/controlsfx/validation/ValidationSupport.java
```diff
@@ -158,6 +158,24 @@ public void redecorate() {
...
         }
     }
+
+    /**
+     * Triggers validation for all known components.
+     * It is only necessary to call this if it is needed to revalidate even if the value of the control has not changed.
+     */
+    public void revalidate() {
+        for (Control target : getRegisteredControls()) {
+            target.fireEvent(new ValidateEvent());
```
eugener, line 168: I wonder why ValidationEvent instance has to be created every time? 
It does not cary any specific information. I'd think singleton instance should be enough.

controlsfx/src/main/java/org/controlsfx/validation/ValidationSupport.java
```diff
@@ -158,6 +158,24 @@ public void redecorate() {
...
+        }
+    }
+
+    /**
+     * Triggers validation for the given component.
+     * It is only necessary to call this if it is needed to revalidate even if the value of the control has not changed.
+     */
+    public void revalidate(Control c) {
+        c.fireEvent(new ValidateEvent());
```
eugener, line 177: Same here.... singleton instance can be used

controlsfx/src/main/java/org/controlsfx/validation/ValidationSupport.java
shollander: Event objects are not meant to be reused. If you look at the parent `javafx.event.Event` class it does carry state such as the `consumed` field. Logically also, it is a different occurrence of the event, so it should be a different object.

eugener: Not a big deal... was simply wondering if optimization is possible.

Issue #812, opened 2016-02-03
ToggleSwitch thumb becomes permanently misaligned


JonathanGiles:
**[Original report](https://bitbucket.org/controlsfx/controlsfx/issue/621) by Rich Meyer (Bitbucket: [richmeyer7](https://bitbucket.org/richmeyer7), GitHub: [richmeyer7](https://github.com/richmeyer7)).**

----------------------------------------

This bug is for version 8.40.10, which is not available in the Version drop-down.

The ToggleSwitch thumb becomes permanently horizontally misaligned with the thumb area if ToggleSwitchSkin.layoutChildren(double,double,double,double) is called after calling ToggleSwitchSkin.selectedStateChanged() for a transition to the selected state.

This is because the TranslateTransition in ToggleSwitchSkin.selectedStateChanged() applies a non-zero translateX to the thumb and leaves it there, and then layoutChildren() calls setLayoutX(double) to position the thumb without taking into account the thumb's existing non-zero translateX. Thus, method layoutChildren() passes an incorrect value to setLayoutX().

Afterward, the X position is permanently corrupted because method selectedStateChanged() uses byX instead of toX.

You can duplicate the bug by displaying a Scene containing a ToggleSwitch, clicking it to the selected state, and then changing any Scene graph value that will cause a layout pass to call ToggleSwitchSkin.layoutChildren(). Our project has root node pseudo-class state changes for colors and base font size that cause a layout pass.

After you do this, the thumb will be permanently horizontally misaligned.

Another defect in selectedStateChanged() is that more than one TranslateTransition could play simultaneously, if the animation time is set long enough and the user clicks quickly enough (system hysteresis allowing). The second TranslateTransition would get an intermediate translateX as its start value because the first one was still playing, then apply byX and compute an incorrect end value.

As a temporary fix in our project, I extended ToggleSwitchSkin and overrode layoutChildren() to zero out the thumb's translateX after calling the superclass implementation.

A real fix would need to:
1. Handle layoutChildren() colliding with a TranslateTransition (which I think can happen if timing is unlucky).
2. Prevent selectedStateChanged() from assuming the thumb starts in its fully selected or deselected translateX position, and/or prevent the user from triggering multiple simultaneous TranslateTransitions.

[...]

PR #1316, opened 2020-10-16
Fix ToggleSwitch animation when selected

devmagix:
When a ToggleSwitch is shown already selected, then the user unselects
it, the thumb animation does not fire properly. This commit fixes that.
Tested the case where layoutChildren() is called when transition is
active(#812), by changing the font size, works fine.

### DIFF
--- a/controlsfx/src/main/java/impl/org/controlsfx/skin/ToggleSwitchSkin.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/skin/ToggleSwitchSkin.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2015, 2016 ControlsFX
+ * Copyright (c) 2015, 2016, 2020 ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


devmagix:
Opened #1317 to track the issue.

controlsfx/src/main/java/impl/org/controlsfx/skin/ToggleSwitchSkin.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2015, 2016 ControlsFX
+ * Copyright (c) 2015, 2016, 2020 ControlsFX
```
abhinayagarwal, line 2: Update to Copyright (c) 2015, 2020, ControlsFX

controlsfx/src/main/java/impl/org/controlsfx/skin/ToggleSwitchSkin.java
```diff
@@ -88,26 +90,18 @@ public ToggleSwitchSkin(ToggleSwitch control) {
...
         } else {
-            thumb.setTranslateX(thumbArea.getLayoutX());
-            transition.setFromX(0);
-            transition.setToX(distance);
+            // If we are not selected, we need to go from right to left.
+            transition.setRate(-1.0);
+            transition.jumpTo(transition.getDuration());
         }
-        transition.setCycleCount(1);
+
```
abhinayagarwal, line 104: Remove blank line

abhinayagarwal (approved):

PR #1338, opened 2021-02-09
Remove VirtualFlow reflection methods in gridViewSkin

### LINKED ISSUE #1332: GridCell uses protected methods from VirtualFlow
`GridViewSkin` needs to access VirtualFlow#rebuildCells and VirtualFlow#reconfigureCells, which are protected methods and not accessible to GridViewSkin.

We need to dig and find out if these two calls are still necessary for the GridView to work. The code is a bit old so maybe it's not useful anymore.

If it's still useful, an approach similar to spreadsheet's GridViewSkin can be taken in GridView i.e. override the `createVirtualFlow()` method in order to provide a custom VirtualFlow that would either open these two methods, or help accessing them.

Maxoudela:
Fix #1332 

I have ran this on JDK 13 using JavaFX 13. I tried to go above (14 and 15) but gradle went crazy with weird errors. 

I have replaced the call to the VirtualFlow with call to a custom VirtualFlow. I really doubt the methods are still needed but I didn't want to break everything.

Let me know if this is good, and if something else is preventing us from releasing a new version of ControlsFX.


abhinayagarwal:
Tested this from JavaFX 13-15. Works great!

Instead of asking for changes, I have made the changes and pushed the same. Hope you don't mind :)

abhinayagarwal (approved):

Maxoudela:
No worry at all! Thanks!

PR #1341, opened 2021-02-14
Remove VersionChecker

### LINKED ISSUE #1240: Version Checker Seems to Think Java Version is Wrong
I get 

> java.lang.RuntimeException: ControlsFX Error: ControlsFX 11.0.1 requires at least Java Version 9

When I run `doVersionCheck()` from VersionChecker, or I try to use SegmentedButton. My use of ControlsFX is sparse in my project; I only use SegmentedButton, ToggleSwitch, Glyphs, and  AutoCompletionBindings. Seems to only trigger when using the SegmentedButton.

What's confusing me is that I'm targeting OpenJDK 11, my OpenJFX version is set to 13-ea+8, and when I call `System.getProperty("javafx.version")` (which is what VersionChecker does) I get 10.0.2-internal. I want to say this an issue but I haven't seen anyone else complain about the version checker recently. Is the problem on my end?

Also: I'm also getting `WARNING: An illegal reflective access operation has occurred` when I start my application if that helps at all.

abhinayagarwal:
@eugener Since you had made some changes to VersionChecker, you might have a better idea on its usability. I recently removed `controlsfx-build.properties`. Do you think we can get rid of this file as well?

Fixes #1240


eugener:
I think it is a good idea.... if I remember correctly, VersionCheck was done at the time where there were incompatibilities  with specific JavaFX versions, which is not a problem anymore?

abhinayagarwal:
GridView, SpreadsheetView and TableView2 do not work with JavaFX 11.

For now if you try to use ControlsFX 11.x.x with JavaFX 11, you will get either `NoSuchMethodError` or `IllegalAccessException` depending on how ControlsFX is trying to access fields/methods from related classes. 

This is not ideal, but it helps a single version of ControlsFX work with JavaFX 11 through latest JavaFX version. If we can somehow show a better message using VersionChecker, the class will regain it's meaning. Otherwise, we can just remove it.

Maxoudela (approved):

abhinayagarwal:
Hi @eugener ,

Can we merge this PR? Later, if the requirement arise we can always bring the class back.

eugener (approved):
Sorry, my fault. 
Did not approve on time!

PR #1352, opened 2021-03-11
FadeTransition should be based on clear button visibility property and TextField text property

### LINKED ISSUE #1351: Button of clearable TextField not shown
Normally the clearable TextField shows the x-button as long as it contains any text.

However, if the text is changed in rapid succession the closeable TextField can lose track and hide the x-button even if the TextField is not empty.

Here is an example demonstrating the issue:
(the Button toggles the x-button, the textField remains non-empty)

```java
package com.onexip.wago.cleverscript.sheets.datalink;

import javafx.scene.Scene;
import javafx.scene.control.Button;
import javafx.scene.control.TextField;
import javafx.scene.layout.HBox;
import javafx.stage.Stage;
import org.controlsfx.control.textfield.TextFields;

public class ClearableTextfield extends Application
{
    @Override
    public void start(Stage stage) throws Exception
    {
        TextField textField = TextFields.createClearableTextField();
        textField.setText("Hello");

        Button button = new Button("Test");
        button.setOnAction(e -> {
            textField.setText("");
            textField.setText("There!");
        });

        stage.setScene(new Scene(new HBox(textField, button)));
        stage.show();
    }

    public static void main(String[] args)
    {
        launch(args);
    }
}
```

abhinayagarwal:
Fixes #1351

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/textfield/TextFields.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/textfield/TextFields.java
@@ -32,7 +32,6 @@ import impl.org.controlsfx.autocompletion.SuggestionProvider;
 import java.util.Arrays;
 import java.util.Collection;
 
-import javafx.animation.Animation;
 import javafx.animation.FadeTransition;
 import javafx.beans.InvalidationListener;
 import javafx.beans.Observable;
@@ -46,7 +45,6 @@ import javafx.scene.layout.StackPane;
 import javafx.util.Callback;
 import javafx.util.Duration;
 import javafx.util.StringConverter;
-
 import org.controlsfx.control.textfield.AutoCompletionBinding.ISuggestionRequest;
 
 /**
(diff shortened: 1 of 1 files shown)


controlsfx/src/main/java/org/controlsfx/control/textfield/TextFields.java
```diff
@@ -109,6 +110,10 @@ private static void setupClearButtonField(TextField inputField, ObjectProperty<N
 
         inputField.textProperty().addListener(new InvalidationListener() {
             @Override public void invalidated(Observable arg0) {
+                if (fader.getStatus() == Animation.Status.RUNNING) {
+                    fader.stop();
+                    fader.getNode().setOpacity(fader.getToValue());
+                }
```
goldfischli, line 116: Why checking the fader state in the first place?
You could store the button's visible state in a boolean field and use that to switch the state and update the fader.

abhinayagarwal: Hi @goldfischli

Can you add a comment with the suggested code changes ?

goldfischli: I was thinking about something like this:
```java
        inputField.textProperty().addListener(new InvalidationListener() {
            private boolean isButtonVisible = false;

            @Override public void invalidated(Observable arg0) {
                String text = inputField.getText();
                boolean isTextEmpty = text == null || text.isEmpty();

                if (isTextEmpty == isButtonVisible)
                {
                    isButtonVisible = !isTextEmpty;
                    fadeTo(isButtonVisible);
                }
            }

            private void fadeTo( boolean visible ) {
                fader.stop();
                fader.setFromValue(visible? 0.0: 1.0);
                fader.setToValue(visible? 1.0: 0.0);
                fader.play();
            }
        });
```
What do you think about it?

abhinayagarwal: A good solution indeed. I will update the PR.

Maxoudela (approved):

PR #1359, opened 2021-04-05
Add default style-class in SearchableComboBox

### LINKED ISSUE #1358: Add styleclass to SearchableComboBox
SearchableComboBox control doesn't have any default style-class which makes it difficult to style the control without affecting other ComboBox in the same scene.

abhinayagarwal:
Fixes #1358

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/SearchableComboBox.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/SearchableComboBox.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019, ControlsFX
+ * Copyright (c) 2019, 2021, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


abhinayagarwal:
@danielpeintner Can you review this ?

danielpeintner:
> @danielpeintner Can you review this ?

I do not have review rights..  ;-)

Anyhow, I looked at the change and I tried the following CSS styling

```
.combo-box .list-cell {-fx-background: yellow;}
.searchable-combo-box .list-cell {-fx-background: blue;} 
```

The results show as expected blue entries in searchable combo-boxes and yellow entries in *normal* combo-boxes.

Having said that, reversing it to 

```
.searchable-combo-box .list-cell {-fx-background: blue;} 
.combo-box .list-cell {-fx-background: yellow;}
```

does not work as I would expect. Both entries are yellow.

This might be due to CSS rules where I am not super familiar with..

abhinayagarwal:
Hi @danielpeintner ,

Thanks for the review. Your test results are perfectly normal. The SearchableComboBox is actually a ComboBox and any style-class applied to ComboBox should be applied to the SearchableComboBox as well.

danielpeintner (approved):
LGTM

PR #1362, opened 2021-04-16
TableView2: Check cell visibility before add/remove listener

### LINKED ISSUE #1361: Not responsive gui when having many updates in a TableView2 and changing the selected row
When items are updated in a high pace by a thread that is not the JavaFX-Application-Thread and changing the selection back and forth, it can happen that the gui is not responsive anymore. Given that there are many columns not in the visible area.

lord41ph4:
Fixes #1361 

The non-responsiveness came from continuous listener adding while creating the cell that is not visible and be garbage collected immediately. Moving the getCell(..) call further down to be performed after the visibility check is avoiding unnessesary creation of elements.

### DIFF
--- a/controlsfx/src/main/java/impl/org/controlsfx/tableview2/TableRow2Skin.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/tableview2/TableRow2Skin.java
@@ -140,7 +140,7 @@ public class TableRow2Skin<S> extends CellSkinBase<TableRow<S>> {
         getChildren().removeIf(n -> n.getId() != null && n.getId().equals("pane-fixed-cell"));
             
         Object o = control.getProperties().get("fixed");
-        boolean fixedRow = o != null && ((Boolean) o).equals(Boolean.TRUE);
+        boolean fixedRow = Boolean.TRUE.equals(o);
 
         final List<? extends TableColumn<S, ?>> columns = tableView.getVisibleLeafColumns();
 
@@ -168,7 +168,7 @@ public class TableRow2Skin<S> extends CellSkinBase<TableRow<S>> {
         control.verticalShift.setValue(tableView.isRowFixingEnabled() ? getFixedRowShift(index) : 0);
 
         double fixedColumnWidth = 0;
-        List<TableCell<S, ?>> fixedCells = new ArrayList();
+        List<TableCell<S, ?>> fixedCells = new ArrayList<>();
 
         //We compute the cells here
         putCellsInCache();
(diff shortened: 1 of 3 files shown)


abhinayagarwal:
Hi @lord41ph4 ,

Can you add a Test to verify this change?

lord41ph4:
Will do as soon I have time. :)

Maxoudela:
Could you point out where the notable changes are? It seems that the diff is quite large because of a change in indenting.

Also, is this issue reproduced in the SpreadsheetView?

lord41ph4:
Hi @Maxoudela,
I did not change that much actually, it just looks much because I could get rid of the one if statement in the loop that was pointless, because the check was always true because of the continue couple of lines above, and thats why the formatting made the change that big. It is easier when you look the 2 commits (26e78f2 & 8339de5) separately, as the others just added unit tests for different cases. And as I personally like Hamcrest in my unit tests I added the dependency as well. The SpreadsheetView I did not try, and also did not look into its code. I don't know if that is enough explanation, if needed do not hesitate to ask for more. :D

controlsfx/src/test/java/org/controlsfx/control/tableview2/TableView2Test.java
```diff
@@ -0,0 +1,303 @@
...
+        FxToolkit.cleanupStages();
+    }
+
+    @Test
+    public void shouldNotFreeze_When_SelectionIsChangedUnderLoad() {
+        fillTableData();
+        produceUpdateLoad();
+
+        Duration timeTaken = measure(this::changeSelection);
+        assertThat(timeTaken, is(lessThan(MAX_DURATION_FOR_SELECTION_CHANGES)));
```
abhinayagarwal, line 184: What is the logic behind the time taken by `changeSelection` always being less than 20s?

controlsfx/src/test/java/org/controlsfx/control/tableview2/TableView2Test.java
```diff
@@ -0,0 +1,303 @@
...
+    public void shouldReuseItems_When_ScrollingVertically() {
+        fillTableData();
+        measure(() -> scrollDown(60));
+        numberRowChildrenModifications.set(0);
+        measure(() -> scrollUp(60));
+        assertThat(numberRowChildrenModifications.get(), is(0L));
+    }
+
+    @Test
+    public void shouldHaveNoArtifacts_When_ScrollingVertically() {
```
abhinayagarwal, line 206: This test passes even when the fix is not applied. Not sure if its supposed to.

controlsfx/src/test/java/org/controlsfx/control/tableview2/TableView2Test.java
```diff
@@ -0,0 +1,303 @@
...
+    public void shouldNotFreeze_When_SelectionIsChangedUnderLoad() {
+        fillTableData();
+        produceUpdateLoad();
+
+        Duration timeTaken = measure(this::changeSelection);
+        assertThat(timeTaken, is(lessThan(MAX_DURATION_FOR_SELECTION_CHANGES)));
+    }
+
+    @Test
+    public void shouldNotFreeze_When_ScrollingVerticallyUnderLoad() {
```
abhinayagarwal, line 188: This test passes even when the fix is not applied. Not sure if its supposed to.

controlsfx/src/main/java/impl/org/controlsfx/tableview2/TableRow2Skin.java
```diff
@@ -245,174 +246,170 @@ private void requestCellUpdate() {
...
-//                	 tableCell.toFront();
+//                	tableCell.toFront();
                     fixedColumnWidth += width;
-//                    isVisible = true; // If in fixedColumn, it's obviously visible
+//                  isVisible = true; // If in fixedColumn, it's obviously visible
                     fixedCells.add(tableCell);
                 }
             }
-            
-            if (isVisible) {
```
abhinayagarwal, line 255: Why is this check removed?

lord41ph4: Because of line 213 the visible variable was alway true at this point.

controlsfx/src/test/java/org/controlsfx/control/tableview2/TableView2Test.java
lord41ph4: Yes, I gave a maximum of 1 second for each selection change (defined in line 70).

controlsfx/src/test/java/org/controlsfx/control/tableview2/TableView2Test.java
lord41ph4: Yes, it is supposed to be that way. I played around with the code and I thought this test could at least guarantee that the scrolling is not broken. I think it does not harm to have such tests, as they make the life easier when modifying the code

controlsfx/src/test/java/org/controlsfx/control/tableview2/TableView2Test.java
lord41ph4: Yes, it is also supposed to be that way. I first thought I could make the GUI freeze with this test, because I could see that the cells have not been reused before (8339de5). But I guess to make it freeze I need to use heavier cells, like with buttons or something like that. But as it didn't really freeze and the test was anyway written I just left it there.

Maxoudela (approved):

Maxoudela:
I'll modify the SpreadsheetView later on to remove the `isVisible` condition since it's indeed always true after the check with the `continue` is done.

The problem should not appear in the SpreadsheetView since the `getCell` call is done after the `isVisible` check in `GridRowSkin`

PR #1363, opened 2021-05-18
Fix notifications

FlorianKirmaier:
Simplified various places in the code.
I think the minHeight is not required, the content should provide the minimum height.
The computeMinHeight/Width methods caused issues. They also were illogical, because graphic is only a small part of it's content.

After this change it seems to work again.

Without this change, and when the graphic is set (for example with .graphic(new Label(text)) )
Then it looked like the following:

Screenshot 2021-05-18 at 10 55 31

I've also removed the VPos logic, because it now always positions it in the center, and it also doesn't really make sense to me.

### DIFF
--- a/controlsfx/src/main/java/impl/org/controlsfx/skin/NotificationBar.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/skin/NotificationBar.java
@@ -152,10 +152,8 @@ public abstract class NotificationBar extends Region {
         closeBtn.setFocusTraversable(false);
         closeBtn.opacityProperty().bind(transition);
         GridPane.setMargin(closeBtn, new Insets(0, 0, 0, 8));
-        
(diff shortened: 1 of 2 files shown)


abhinayagarwal:
Hi @FlorianKirmaier ,

Can you please sign the CLA as suggested by the bot?

FlorianKirmaier:
Done!

Maxoudela (approved):

Maxoudela:
Can you update the copyright to 
`Copyright (c) 2014, 2021 ControlsFX` please

FlorianKirmaier:
I've updated the copyright dates!

FlorianKirmaier:
The CLA bot seems to be buggy.

Maxoudela:
Indeed

Maxoudela (approved):

abhinayagarwal (approved):

PR #1369, opened 2021-05-31
Action.setSelected(boolean) now triggers the EventHandler if needed,

fmichel:
Hi,

Using org.controlsfx.control.action.Action.setSelected(boolean) changes the selected state of ActionCheck actions correctly but does not trigger the corresponding EventHandler. 

So, changing the state programmatically, using this method, does not make the application do what it would when clicking  toogleButton or checkMenu elements created with the ActionCheck (activate or deactivate the rendering of an animation for instance)

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/action/Action.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/action/Action.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013, 2015, ControlsFX
+ * Copyright (c) 2013, 2021 ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


eugener (approved):
LGTM

controlsfx/src/main/java/org/controlsfx/control/action/Action.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013, 2015, ControlsFX
+ * Copyright (c) 2013, 2015, 2021 ControlsFX
```
abhinayagarwal, line 2: Please update this to `Copyright (c) 2013, 2021, ControlsFX`

controlsfx/src/main/java/org/controlsfx/control/action/Action.java
```diff
@@ -190,7 +190,12 @@ public final boolean isSelected() {
      * @param selected
      */
     public final void setSelected( boolean selected ) {
-    	selectedProperty.set(selected);
+    	if (isSelected() != selected) {
+    		selectedProperty.set(selected);
+    		if(getClass().isAnnotationPresent(ActionCheck.class)) {
+    			eventHandler.accept(null);
+    		}
+    	}
```
abhinayagarwal, line 198: Please use SPACE instead of TAB

fmichel: I did the requested changes

Issue #1275, opened 2020-05-20
NumericField in PropertySheet does not fire a change when set to zero for the first time


bezvaapp:
controlsfx 12.0.0-SNAPSHOT
When a NumericField is created, the initial value of the value filed is set to 0. During the initialization only the text of the TextField is set. Therefore when a user enters 0 for the first time, the NumericField does not identifies it as a change and the listeners are not called. 

It may be related to #66.

The user first must enter a non-zero value or click the field to get the focus then leave the filed. When the NumericField loses the focus finally it sets the value to zero.

The problem is in this method. It sets only the text and not the actual value which remains 0.
`Editors.createNumericEditor(Item property )`
```
            @Override public void setValue(Number value) {
                sourceClass = (Class<? extends Number>) value.getClass();
                getEditor().setText(value.toString());
            }
```
You can test it. When you enter 0 in the field for the first time, the 0 is not propagated into the bean.

```
public class PropertySheetApp  extends Application {

	public static void main(String[] args) {
		launch(args);
	}

	@Override
	public void start(Stage stage) {
		
		var bean = new Bean();
		bean.setNumber(10);
		var items = BeanPropertyUtils.getProperties(bean);
		var sheet = new PropertySheet(items);
		
		StackPane pane = new StackPane(sheet);

		stage.setScene(new Scene(pane, 300, 250));
		stage.show();
	}

	public static class Bean {

		private double number;
		
		public double getNumber() {
			return number;
		}
		
		public void setNumber(double number) {
			System.out.println(number);
			this.number = number;
		}
	}
}
```

fsimons:
In ControlsFX 11.1.1 the issue is somehow fixed. However, I think the behaviour is still not perfect. This can be tested with the example given above. If I enter 0 or any other number the first time, the change event is fired. However, if I enter the same number again, there is no change event. This is somehow correct, however the number is not formatted correctly anymore.

PR #1371, opened 2021-06-24
Fix and unit test for issue #1275

SamCooper:
Hello, after hitting this issue in Property Sheets I thought I could take the time to create the fix based on the suggestion in issue #1275 and create the unit test for it too.

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/property/editor/Editors.java
+++ b/controlsfx/src/main/java/org/controlsfx/property/editor/Editors.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2015 ControlsFX
+ * Copyright (c) 2015, 2021, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/controlsfx/src/main/java/org/controlsfx/property/editor/Editors.java
+++ b/controlsfx/src/main/java/org/controlsfx/property/editor/Editors.java
@@ -98,6 +98,7 @@ public class Editors {
 
             @Override public void setValue(Number value) {
                 sourceClass = (Class<? extends Number>) value.getClass();
+                ((NumericField)getEditor()).setValue(value);
                 getEditor().setText(value.toString());
             }
             
--- a/controlsfx/src/main/java/org/controlsfx/property/editor/Editors.java
+++ b/controlsfx/src/main/java/org/controlsfx/property/editor/Editors.java
@@ -98,7 +98,7 @@ public class Editors {
 
             @Override public void setValue(Number value) {
                 sourceClass = (Class<? extends Number>) value.getClass();
-                ((NumericField)getEditor()).setValue(value);
+                getEditor().setValue(value);
                 getEditor().setText(value.toString());
             }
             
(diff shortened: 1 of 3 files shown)


controlsfx/src/main/java/org/controlsfx/property/editor/Editors.java
```diff
@@ -98,6 +98,7 @@ private Editors() {
 
             @Override public void setValue(Number value) {
                 sourceClass = (Class<? extends Number>) value.getClass();
+                ((NumericField)getEditor()).setValue(value);
```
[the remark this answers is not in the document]
eugener: Would it be better to listen for text changes within the `NumericField` and set the value internally on text change without exposing `value`.

controlsfx/src/test/java/org/controlsfx/property/editor/NumericPropertyZeroTest.java
```diff
@@ -0,0 +1,137 @@
+/**
+ * Copyright (c) 20121, ControlsFX
```
Maxoudela, line 2: Typo

controlsfx/src/main/java/org/controlsfx/property/editor/NumericField.java
```diff
@@ -77,6 +77,10 @@ public NumericField( Class<? extends Number> cls ) {
         }
     }
 
+    protected void setValue(Number v) {
```
Maxoudela, line 80: If we do keep this method (see @eugener comment), it will need a javadoc.

Maxoudela (changes requested):

controlsfx/src/main/java/org/controlsfx/property/editor/Editors.java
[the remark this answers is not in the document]
SamCooper: I don't think you do need to check the cast as you've passed it into AbstractPropertyEditor, and set the generic type of AbstractPropertyEditor to NumericField.

If fact the code at that line should just be:

`getEditor().setValue(value);`

without the cast (compiles and works fine actually).

I tried the listener approach and it occasionally resulted in `java.lang.IllegalArgumentException: The start must be <= the end` I think caused by an empty string. It was difficult to replicate as some sort of race condition was happening.




Anyway, TLDR, I think the cast can be removed and all is ok. I'll push an update to reflect this.

controlsfx/src/test/java/org/controlsfx/property/editor/NumericPropertyZeroTest.java
SamCooper: Fixed

controlsfx/src/main/java/org/controlsfx/property/editor/NumericField.java
SamCooper: Fixed

Maxoudela (approved):

abhinayagarwal:
Hi @SamCooper,

Can you please update the license header in the 2 files changed to "Copyright (c) 2015, 2021, ControlsFX" ?

eugener (approved):

abhinayagarwal (approved):

PR #1385, opened 2021-08-23
#1384: Fix using active popups as owners for new popups

### LINKED ISSUE #1384: Some notifications disappearing too early when application has no active window
Without any active application window and multiple notifications showing, all will disappear at the same time as the oldest one. This happens because active notifications are taken an owner for new notifications. When owner disappears - child popups also disappear. Popup owner is set here:
https://github.com/controlsfx/controlsfx/blob/jfx-13/controlsfx/src/main/java/org/controlsfx/control/Notifications.java#L501
`ownerWindow` value from here:
https://github.com/controlsfx/controlsfx/blob/jfx-13/controlsfx/src/main/java/org/controlsfx/control/Notifications.java#L378-L385
could be used instead to avoid the issue.

dawidm:
fixes #1384

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/Notifications.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/Notifications.java
@@ -379,6 +379,8 @@ public class Notifications {
             // Stylesheets which are added to the scene of a popup aren't
             // considered for styling. For this reason, we need to find the next
             // window in the hierarchy which isn't a popup.
+            // This is also used as an owner for a new popup
+            // to avoid using previous popups as owners.
             Window ownerWindow = owner;
             while (ownerWindow instanceof PopupWindow) {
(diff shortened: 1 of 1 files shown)


Maxoudela (approved):

Maxoudela:
Can you sign the CLA @dawidm ?

dawidm:
> Can you sign the CLA @dawidm ?

Yes, I just did it.

abhinayagarwal:
Hi @dawidm ,

Is it possible to test this change via a Test?

dawidm:
> Hi @dawidm ,
> 
> Is it possible to test this change via a Junit Test?

Hi, I don't think so.

abhinayagarwal:
OK, I will merge the PR as @Maxoudela has already approved it.

Issue #713, opened 2015-04-23
NotificationPane should not be focusTraversable


JonathanGiles:
**[Original report](https://bitbucket.org/controlsfx/controlsfx/issue/522) by Robert Lichtenberger (Bitbucket: [effad](https://bitbucket.org/effad), GitHub: [effad](https://github.com/effad)).**

----------------------------------------

Steps to reproduce:
* Start controlsfx-samples
* Click on *Notification Pane*, then on *Sample* tab
* The focus is now on *Sample*
* Press <Tab>
* The focus is now on the Notification Pane, without any visual affordance.
* Press <Tab> again
* The focus is now on the *Show/Hide* button.

=> NotificationPane should not be focusTraversable.

PR #1395, opened 2021-09-19
Remove behavior from RangeSlider and SnapshotView control

### LINKED ISSUE #1073: Need various --add-exports to have RangeSlider work in modularized project.
**[Original report](https://bitbucket.org/controlsfx/controlsfx/issue/882) by Sebastien Kister (Bitbucket: [SebKister](https://bitbucket.org/SebKister), GitHub: [SebKister](https://github.com/SebKister)).**

----------------------------------------

I added a single RangeSlider control to a modularized Java project under Oracle JDK 10.0.2.
In order to run the application I had to add the following to the VM arguments:

--add-exports=javafx.base/com.sun.javafx.runtime=controlsfx
--add-exports=javafx.controls/com.sun.javafx.scene.control.behavior=controlsfx
--add-exports=javafx.controls/com.sun.javafx.scene.control.inputmap=controlsfx
--add-exports=javafx.graphics/com.sun.javafx.scene.traversal=controlsfx

abhinayagarwal:
Fixes #1073


controlsfx/src/main/java/impl/org/controlsfx/skin/RangeSliderSkin.java
```diff
@@ -142,290 +126,197 @@ public RangeSliderSkin(final RangeSlider rangeSlider) {
...
-                if (hasFocus) {
-                    lowThumb.setFocus(true);
-                } else {
-                    lowThumb.setFocus(false);
-                    highThumb.setFocus(false);
-                    currentFocus = NONE;
-                }
+        rangeSlider.focusedProperty().addListener((ov, t, hasFocus) -> {
+            if (hasFocus) {
+                lowThumb.setFocus(true);
```
DaveJarvis, line 147: Also consider:

```
lowThumb.setFocus(hasFocus);

if(!hasFocus) {
    highTumb.setFocus(false);
    currentFocus = NONE;
}
```

controlsfx/src/main/java/impl/org/controlsfx/skin/RangeSliderSkin.java
```diff
@@ -142,290 +126,197 @@ public RangeSliderSkin(final RangeSlider rangeSlider) {
...
+            event.consume();
+        };
+        getSkinnable().addEventHandler(KeyEvent.KEY_PRESSED, keyPressEventHandler);
+        getSkinnable().addEventHandler(KeyEvent.KEY_RELEASED, keyReleaseEventHandler);
+        // set up a callback to indicate which thumb is currently selected (via enum).
+        setSelectedValue(v -> currentFocus);
+    }
+    
+    private void focusPreviousChild() {
+        final int index = getSkinnable().getParent().getChildrenUnmodifiable().indexOf(getSkinnable());
```
DaveJarvis: Consider caching the unmodifiable list in a local variable (for reuse) to ensure that the list reference doesn't change between calls to getParent.

controlsfx/src/main/java/impl/org/controlsfx/skin/RangeSliderSkin.java
```diff
@@ -142,290 +126,197 @@ public RangeSliderSkin(final RangeSlider rangeSlider) {
...
+        if (index != 0) {
+            getSkinnable().getParent().getChildrenUnmodifiable().get(index - 1).requestFocus();
+        } else {
+            getSkinnable().getParent().requestFocus();
+        }
+    }
+
+    private void focusNextChild() {
+        final int index = getSkinnable().getParent().getChildrenUnmodifiable().indexOf(getSkinnable());
```
DaveJarvis: Would it be safer to assign getChildrenUnmodifiable to a local `final var` for reuse?

controlsfx/src/main/java/impl/org/controlsfx/ImplUtils.java
```diff
@@ -157,4 +157,22 @@ public static void stripRootPane(Scene scene, Parent originalParent, boolean use
 
         return children;
     }
+
+    public static void focusPreviousSibling(Node node) {
+        final int index = node.getParent().getChildrenUnmodifiable().indexOf(node);
+        if (index != 0) {
+            node.getParent().getChildrenUnmodifiable().get(index - 1).requestFocus();
```
Maxoudela, line 164: If `indexOf` returns `-1`, won't this line tries to retrieve a children at index `-2` ?

controlsfx-samples/src/main/java/org/controlsfx/samples/HelloNotificationPane.java
```diff
@@ -26,8 +26,6 @@
  */
```
Maxoudela, line 26: The copyright needs to be updated here.

Maxoudela (changes requested):

controlsfx/src/main/java/impl/org/controlsfx/skin/NotificationPaneSkin.java
```diff
@@ -112,20 +113,27 @@ public NotificationPaneSkin(final NotificationPane control) {
...
+                if (getChildren().size() > 0) {
+                    getChildren().get(0).requestFocus();
+                }
+            }
+        });
 
-        // Fix for Issue #522: Prevent NotificationPane from receiving focus
-        ParentTraversalEngine engine = new ParentTraversalEngine(getSkinnable());
-        ReflectionUtils.setTraversalEngine(control, engine);
-        engine.setOverriddenFocusTraversability(false);
```
abhinayagarwal, line 119: Adding a fix for these 3 lines has made me realized the importance of a Traversal Engine in JavaFX. Most of the heavy lifting was already done by Jonathan in [JDK-8091673](https://bugs.openjdk.java.net/browse/JDK-8091673). @Maxoudela @eugener Do you think its worth pursuing this?

abhinayagarwal: AFAIU, it disables the Parent nodes in a custom control from receiving focus. It was implemented as a fix for https://github.com/controlsfx/controlsfx/issues/713  by @effad

effad: I can confirm what @abhinayagarwal has said about these lines. The issue-number #522 refers to BitBucket-Issues (no longer existant ...).
I also would appreciate if the TraversalEngine would be part of the public API. After all, the above lines are very hacky :-).

controlsfx/src/main/java/impl/org/controlsfx/skin/NotificationBar.java
```diff
@@ -138,54 +137,83 @@ public NotificationBar() {
...
+        };
+        addEventHandler(KeyEvent.KEY_PRESSED, keyPressEventHandler);
     }
 
     void updatePane() {
         actionsBar = ActionUtils.createButtonBar(getActions());
         actionsBar.opacityProperty().bind(transition);
+        actionsBar.focusedProperty().addListener((o, ov, hasFocus) -> {
+            if (!actionsBar.getButtons().isEmpty()) {
+                actionsBar.getButtons().get(0).requestFocus();
```
DaveJarvis, line 176: Do we care about thread-safety? If so, then it's possible that the button is removed between lines 175 and 176, resulting in a `NullPointerException`. Thoughts on an OOP alternative that upholds encapsulation? Maybe it's not possible, but something like:

```
actionBar.giveButtonFocus(OK_BUTTON);  // OK_BUTTON constant = 0
```

That would eliminate the need for the conditional (in this method anyway). The `actionsBar` class would need a method like:

```
public void giveButtonFocus( final int index ) {
  final var buttons = getButtons();

  // index is 0-based, size starts at 1 if not empty, so use >
  if( buttons.size() > index ) {
    buttons.get( index ).requestFocus();
  }
}
```

That's then reusable, thread-safe, and will throw neither an `NullPointerException` nor an `ArrayIndexOutOfBounds` exception.

controlsfx/src/main/java/impl/org/controlsfx/skin/RangeSliderSkin.java
abhinayagarwal: I like this refactoring :)

controlsfx/src/main/java/impl/org/controlsfx/skin/RangeSliderSkin.java
abhinayagarwal: Its highly likely that this will happen. But, better safe than sorry ;)

controlsfx/src/main/java/impl/org/controlsfx/ImplUtils.java
abhinayagarwal: May be its better to use `if (index > 0)` ?

controlsfx/src/main/java/impl/org/controlsfx/skin/NotificationBar.java
abhinayagarwal: JavaFX is a single threaded GUI framework i.e. all scene-graph activities are done on the JavaFX application thread. Therefore, activities like removing of nodes, focus etc. are all inherently executed synchronously.

controlsfx/src/main/java/impl/org/controlsfx/ImplUtils.java
abhinayagarwal: Updated PR with this change.

abhinayagarwal:
Marking this PR as RFR

Maxoudela (approved):

PR #1415, opened 2022-02-03
Fixed bug where high thumb & rangeBar missing

qtran:
Hi,
This PR fix a bug where the high thumb and range bar are missing in RangeSlider when you change the tick marks / labels visibility
I think the problem culprit is here: https://github.com/controlsfx/controlsfx/blob/b0cec184fb62395cc51edd39a3790e7589f467a9/controlsfx/src/main/java/impl/org/controlsfx/skin/RangeSliderSkin.java#L314
The method documentation says: "After this method, we must be sure to add the high Thumb and the rangeBar." but I don't see any code line which does that.
It worked fine with 8.40.16 though


Maxoudela:
Can you open an issue in the tracker with a description of the bug so that this PR can be linked to it?

qtran:
Sorry better late than never

PR #1436, opened 2022-05-01
HyperlinkLabel focus property update

### LINKED ISSUE #1321: HyperlinkLabel setFocusTraversable() not working as expected
When setting `setFocusTraversable(false)` on a `HyperlinkLabel` the embedded link node is still focustraversable. One would expect that setting `setFocusTraversable(false)` on the `HyperlinkLabel` would set traversable to false for all nodes contained in the `HyperlinkLabel`.

I use the following versions:
```
<dependency>
   <groupId>org.openjfx</groupId>
   <artifactId>javafx-controls</artifactId>
   <version>15.0.1</version>
</dependency>
<!-- https://mvnrepository.com/artifact/org.controlsfx/controlsfx -->
<dependency>
    <groupId>org.controlsfx</groupId>
    <artifactId>controlsfx</artifactId>
    <version>11.0.3</version>
</dependency>
```

abhinayagarwal:
Fixes #1321

### DIFF
--- a/controlsfx/src/main/java/impl/org/controlsfx/skin/HyperlinkLabelSkin.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/skin/HyperlinkLabelSkin.java
@@ -29,6 +29,7 @@ package impl.org.controlsfx.skin;
 import java.util.ArrayList;
 import java.util.List;
 
+import javafx.beans.value.ChangeListener;
 import javafx.event.ActionEvent;
 import javafx.event.EventHandler;
 import javafx.geometry.Insets;
@@ -70,6 +71,7 @@ public class HyperlinkLabelSkin extends SkinBase<HyperlinkLabel> {
             }
         }
     };
+    private final ChangeListener<Boolean> focusChangeListener;
 
     /***************************************************************************
      * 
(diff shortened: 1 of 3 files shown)


controlsfx/src/main/java/impl/org/controlsfx/skin/HyperlinkLabelSkin.java
```diff
@@ -70,52 +70,59 @@ public class HyperlinkLabelSkin extends SkinBase<HyperlinkLabel> {
...
         super(control);
-        
-        this.textFlow = new TextFlow();
+
+        textFlow = new TextFlow();
+        hyperlink = new Hyperlink();
+        hyperlink.setPadding(new Insets(0, 0, 0, 0));
+        hyperlink.setOnAction(eventHandler);
+        hyperlink.focusTraversableProperty().bind(control.focusTraversableProperty());
+        control.focusedProperty().addListener((o, ov, nv) -> {
```
Maxoudela, line 88: Just a minor concern whether this should be inside a WeakListener or not. I had so many memory leaks in applications because of listener added directly. 
Don't know if this one can cause a memory leak or not

Maxoudela (approved):

controlsfx/src/main/java/impl/org/controlsfx/skin/HyperlinkLabelSkin.java
abhinayagarwal: Valid point. I have overridden `dispose` method to unbind bindings and remove listeners.

abhinayagarwal: I now use `SkinBase#registerChangeListener` which is called and gc'd during dispose automatically. Another round of review is appreciated.

Issue #1441, opened 2022-05-08
Solution for application crash on Window close, when PopOver is still open.


rhabarberbarberah:
The problem seems to be caused by the fade-duration, which will keep the PopOver visible, while the Window already closed.

You can keep your default fade-furation, if you call:
```
ownerWindowProperty().addListener((o) -> {
    if (getOwnerWindow() != null) {
        getOwnerWindow().setOnCloseRequest((t) -> {
            hide(Duration.ZERO);
        });
    }
});
```
instead.

That way it'll only disappear immediately, when the ownerWindow is closed.

_Originally posted by @rhabarberbarberah in https://github.com/controlsfx/controlsfx/issues/448#issuecomment-1120431005_

rhabarberbarberah:
might be worth considering to add this snippet to the PopOver-Class

abhinayagarwal:
Hi @rhabarberbarberah ,

Would you like to create a PR for this issue?

rhabarberbarberah:
Hi, @abhinayagarwal

just created a PR: #1443

PR #1443, opened 2022-05-08
Remove event filter when owner window is hiding 

rhabarberbarberah:
As Described under https://github.com/controlsfx/controlsfx/issues/1441
The Error on Window closing while a PopOver is open is caused by the fadeout time.
Can be fixed by adding a OnCloseRequest-listener to the OwnerWindow which then closes the PopOver without delay.

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/PopOver.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/PopOver.java
@@ -153,8 +153,11 @@ public class PopOver extends PopupControl {
     }
 
     /**
+     * Adds the EventFilter when the WindowProperty changes.
+     * Solves the issue (#1441) that leads to an exception, when closing the window (using the "x"-button) while the PopOver is visible.
      *
-     *
+     * Hint: This seems to be needed additional to filters that are already implemented in {@link #show(Window)}, {@link #hide(Duration)}...
+     * It also doesn't work when using the WeakEventHandler instead. Replacing the Handlers in the other positions also doesn't lead to success.
      */
     private void setUpOwnerCloseBehaviour() {
         ownerWindowProperty().addListener((o) -> {
--- a/controlsfx/src/main/java/org/controlsfx/control/PopOver.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/PopOver.java
@@ -164,9 +164,6 @@ public class PopOver extends PopupControl {
             if (getOwnerWindow() != null) {
                 getOwnerWindow().addEventFilter(WindowEvent.WINDOW_CLOSE_REQUEST, closePopOverOnOwnerWindowCloseLambda);
                 getOwnerWindow().addEventFilter(WindowEvent.WINDOW_HIDING, closePopOverOnOwnerWindowCloseLambda);
-            }else{
-                getOwnerWindow().removeEventFilter(WindowEvent.WINDOW_CLOSE_REQUEST, closePopOverOnOwnerWindowCloseLambda);
-                getOwnerWindow().removeEventFilter(WindowEvent.WINDOW_HIDING, closePopOverOnOwnerWindowCloseLambda);
             }
         });
     }
@@ -534,6 +531,10 @@ public class PopOver extends PopupControl {
                     closePopOverOnOwnerWindowClose);
             ownerWindow.removeEventFilter(WindowEvent.WINDOW_HIDING,
                 closePopOverOnOwnerWindowClose);
+            ownerWindow.removeEventFilter(WindowEvent.WINDOW_CLOSE_REQUEST,
+                    closePopOverOnOwnerWindowCloseLambda);
+            ownerWindow.removeEventFilter(WindowEvent.WINDOW_HIDING,
+                    closePopOverOnOwnerWindowCloseLambda);
         }
         if (fadeOutDuration == null) {
             fadeOutDuration = DEFAULT_FADE_DURATION;
--- a/controlsfx/src/main/java/org/controlsfx/control/PopOver.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/PopOver.java
@@ -115,14 +115,8 @@ public class PopOver extends PopupControl {
             }
         });
 
-        ownerWindowProperty().addListener((o) -> {
-            if (getOwnerWindow() != null) {
-                getOwnerWindow().setOnCloseRequest((t) -> {
-                    hide(Duration.ZERO);
-                });
-            }
-        });
-        
+        setUpOwnerCloseBehaviour();
+
         /*
          * Create some initial content.
          */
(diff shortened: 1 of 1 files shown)


rhabarberbarberah:
I did the CLA-Thing, let me know, if anything is missing

abhinayagarwal:
PopOver already [adds event filter](https://github.com/controlsfx/controlsfx/blob/871c3396f8b0c3ab2444e813bcdd0bc1f8d07df5/controlsfx/src/main/java/org/controlsfx/control/PopOver.java#L469) on owner window to hide without delay on `WindowEvent.WINDOW_CLOSE_REQUEST` and `WindowEvent.WINDOW_HIDING` events. Under what condition will these events fail and your additional listener help?

rhabarberbarberah:
Hi @abhinayagarwal,
I just had a look at the section, you're right, that should -- in theory -- do the job and I missed that in the beginning, since I ofiginally solved the problem from outside the class.
The filter doesn't seem to work as intended.
You should be able to reproduce the behaviour that was originally described in #448 by running the FX-Sample-Application.
Just close the Window while the PopOver is visible and you should see the described exception.

rhabarberbarberah:
Taking a closer look it seems to make a difference, if you close the Window using "alt+F4", in that case the error doesn't seem to happen.

rhabarberbarberah:
further investigating the ownerWindowClosing()-method doesn't seem to be executed, when you close the Window via the "x"-button

rhabarberbarberah:
using the workaround I posted in the PR the error goes away.
Maybe it has something to do with how eventfilters are executed. Didn't dig deeper into that topic but I also noticed unexpected behaviour when playing around with eventfilters/-handlers & -listners in different contexts.
But I'm curious, does that give you a hint on the reason behind the error? maybe a more elegant solution?

rhabarberbarberah:
Using just one Stage it might not seem that relevant to fix the error.
But in my usecase the whole application crashes when closing a subwindow. 
--> that's why I was so keen to find a solution

rhabarberbarberah:
I dug a little deeper, refactored the Case in parallel to my personal usecase.
1. I noticed, that the first try didn't help, when multiple Popovers are open at a time
   - solved by using eventfilter insted

also did some refactoring and cleanup 

Important to note:
The existing implementation adds a WeakEventHandler. This approach doesn't seem to work for this case.
Also tested setups, where I tried out to replace EventHandlers/WeakEventHandlers in various places.
But they either don't solve the issue, or lead to additional exceptions.

abhinayagarwal:
do you have a SSCCE to test this PR?

rhabarberbarberah:
Sure, not quite at hand, but I'm sure I can extract the essential part off my project to reduce it to the essential part.
Will need some time. propably on weekend.

rhabarberbarberah:
Hi,
sorry, got distracted and temporary forgot, about this.
Here's a link to a short example, that should make it easy to test the issue.
Looking forward to your feedback.

rhabarberbarberah:
https://github.com/rhabarberbarberah/PopOverBugFix

rhabarberbarberah:
@abhinayagarwal Does the example help?
Are there any issues?

abhinayagarwal:
I was trying to find the reason why it works with this fix and not with the old fix but haven't been able to. I am missing something.

Nevertheless, this approach works and I am OK to go forward with it. I have the following recommendations:

1. Change the InvalidateListener to ChangeListener. If the old value  in change listener is not null, remove the event handlers from it
2. Remove the following piece of code from all the `show()` methods as they seem to be redundant:

```
ownerWindow.addEventFilter(WindowEvent.WINDOW_CLOSE_REQUEST,
                closePopOverOnOwnerWindowClose);
ownerWindow.addEventFilter(WindowEvent.WINDOW_HIDING,
                closePopOverOnOwnerWindowClose);
```

rhabarberbarberah:
yes, you're right.
On the paper it looks like the old fix should do the job. I've also not been able to figure out, why it fails.

rhabarberbarberah:
Hi,
I just pushed the recommended changes :)

abhinayagarwal:
Ah, this bot needs fixing :p

rhabarberbarberah:
Yea, was just gona ask about that.

still seems to run without exceptions.

While I removed the redundant lines from the show()-methods i noticed, that the ownerWindow field is reassigned in the process.
if the EventFilters would be set BEFORE reassignment, that could a reason, why they sometimes were not triggered. (although it doesn't seem to apply in our case)
I also noticed, that the ownerWindow field shadows a field in the PopupWindow-Class, which is referenced by the ownerWindowProperty(). ->suboptimal, because it makes it harder to understand what is referenced & overridden in different calls/assignments.
Might be connected to why the initial fix didn't always work.

abhinayagarwal:
You are absolutely right about `ownerWindow` field getting re-assigned in `show()`. However, `addFilter` statements are always present at the end i.e. after the re-assignment and call to `super()`, which means that both `ownerWindow` and `getownerWindow()` should (theoretically) have the same reference.

I also tested this by checking the values in `hide()` method, leading to more frustration :)

Weak listeners are notorious in JavaFX and could also be a reason, so I got rid of them and used the listener `closePopOverOnOwnerWindowCloseLambda` in show methods but that didn't change anything :)

If there is a bug, I would want to know what's causing it. Unfortunately, this is a peculiar case :D

abhinayagarwal:
When parent is closed, `hide()` is called before `WINDOW_CLOSE_REQUEST` event is fired. This leads to removal of filters attached to the window events. This has been causing the whole ruckus. Thanks to @jperedadnr for digging it out.

The fix is quite simple, remove the following lines from `hide()`:

```
if (ownerWindow != null){
        ownerWindow.removeEventFilter(WindowEvent.WINDOW_CLOSE_REQUEST,
                    closePopOverOnOwnerWindowClose);
        ownerWindow.removeEventFilter(WindowEvent.WINDOW_HIDING,
                closePopOverOnOwnerWindowClose);
}
```

I have made a [commit](https://github.com/abhinayagarwal/controlsfx/commit/2b8cb9cbf3345a575a8d9e9151ade6f2e2a16fb8) with the fix. Please test it out and let me know if it works.

abhinayagarwal:
Hi @rhabarberbarberah ,

did you get a chance to test it?

rhabarberbarberah:
Hey @abhinayagarwal,

just tested it with the sampler and my test application.
the issue seems to be fixed by this change.

greetings

abhinayagarwal:
Hi @rhabarberbarberah ,

Thank you for the test. I have simplified the fix and pushed it to this PR.

abhinayagarwal (approved):

rhabarberbarberah:
Hi @abhinayagarwal,
this is great news.
looking forward to the next release, then I can remove my workaround from my project.

Issue #1521, opened 2023-10-06
RangeSlider: range bar jumps on dragging when orientation is vertical


james-d:
See the stack overflow question at https://stackoverflow.com/questions/77239808/controlsfx-rangeslider-is-laggy-when-orientation-is-vertical

For some reason, this only reproduces for me when the example application (reproduced below) is running on the Retina display on my MacBook; it does not reproduce when the app is on a (lower-resolution) external monitor.

The following code reproduces the issue. Shorten the range of the vertical range bar and drag the *range bar* (not the thumbs).

```

import javafx.application.Application;
import javafx.geometry.Orientation;
import javafx.geometry.Pos;
import javafx.scene.Scene;
import javafx.scene.layout.*;
import javafx.stage.Stage;
import org.controlsfx.control.RangeSlider;

public class App extends Application {

    @Override
    public void start(Stage stage) {

        RangeSlider rangeSliderVertical = new RangeSlider();
        rangeSliderVertical.setOrientation(Orientation.VERTICAL);
        rangeSliderVertical.setMinHeight(500);
        rangeSliderVertical.setMax(100);
        rangeSliderVertical.setMin(0);
        rangeSliderVertical.setHighValue(100);
        rangeSliderVertical.setLowValue(0);

        rangeSliderVertical.lowValueProperty().addListener((obs, oldValue, newValue) -> System.out.printf("%.2f -> %.2f%n", oldValue.doubleValue(), newValue.doubleValue()));

        RangeSlider rangeSliderHorizontal = new RangeSlider();
        rangeSliderHorizontal.setMinWidth(500);
        rangeSliderHorizontal.setMax(100);
        rangeSliderHorizontal.setMin(0);
        rangeSliderHorizontal.setHighValue(100);
        rangeSliderHorizontal.setLowValue(0);
        rangeSliderHorizontal.setOrientation(Orientation.HORIZONTAL);

        FlowPane vBoxWithSliders = new FlowPane();
        VBox.setVgrow(vBoxWithSliders, Priority.ALWAYS);
        HBox.setHgrow(vBoxWithSliders, Priority.ALWAYS);

        vBoxWithSliders.setAlignment(Pos.CENTER);
        vBoxWithSliders.getChildren().addAll(rangeSliderVertical, rangeSliderHorizontal);

        var scene = new Scene(vBoxWithSliders, 600, 600);
        stage.setScene(scene);
        stage.show();

    }

    public static void main(String[] args) {
        launch();
    }

[...]

PR #1522, opened 2023-10-06
Update RangeSliderSkin as fix for #1521

### LINKED ISSUE #1099: RangeSlider stutters when the track is dragged in vertical orientation.
When the RangeSlider _track_ is dragged there is a visible stutter with both the low and high thumbs rapidly moving to other points on the track and then back again. The distance between the low and high thumb can also change. The effect is reduced by dragging the track very slowly. This does not occur when the RangeSlider is horizontal. Tested on the version 8.14.

james-d:
Fixes jumping of range bar when dragging range bar in vertical range slider.

Fixes #1099


james-d:
I think the first time I did this I had the incorrect GitHub user name. That should be rectified.

Siedlerchr:
Tested it on my Macbook M1 an could reproduce the original issue and also confirm the fix

Siedlerchr (approved):
LGTM. Could confirm the fix on macOS

abhinayagarwal (approved):

PR #1539, opened 2024-01-31
Move border creation outside of layout pass

### LINKED ISSUE #1538: PopOver has infinite layout passes when adding line border with title
Run the following simple test case with current ControlsFX 11.2.0 and JavaFX 21.0.2:

```
import javafx.scene.Node;
import javafx.scene.Scene;
import javafx.scene.control.Button;
import javafx.scene.control.Label;
import javafx.scene.layout.Pane;
import javafx.scene.layout.StackPane;
import javafx.stage.Stage;
import org.controlsfx.control.PopOver;
import org.controlsfx.tools.Borders;

public class PopOverTest extends Application {

    @Override
    public void start(final Stage primaryStage) {
        final Button button = new Button("Button");
        final StackPane root = new StackPane(button);
        button.setOnAction(e -> {
            final PopOver popOver = new PopOver();
            StackPane pane = new StackPane(new Label("Text"));
            Node node = Borders.wrap(pane).lineBorder()
                    .buildAll();
            ((Pane) node).needsLayoutProperty().addListener((obs, ov, nv) -> System.out.println("needs layout = " + nv));
            popOver.setContentNode(node);
            popOver.show(root);
        });
        Scene scene = new Scene(root, 200, 100);
        primaryStage.setScene(scene);
        primaryStage.show();
    }
}
```
Click on the button to show the popOver control and see the output:

image
```
needs layout = false
needs layout = true
needs layout = false
```
That's it (as expected).

Now add a title to the border:
```
            Node node = Borders.wrap(pane).lineBorder()
                     .title("Title")
                    .buildAll();
```
and run again:

image

and check the output:

```
needs layout = false
needs layout = true
needs layout = false
needs layout = true
needs layout = false
needs layout = true
needs layout = false
needs layout = true
needs layout = false
needs layout = true
...
```
it doesn't stop while the popup is showing up.

(This can be reproduced on all platforms)

jperedadnr:
Fixes #1538 

This PR modifies the `Borders#StrokeBorder` class to avoid doing node creation on layout passes, by moving out the border creation to a method that is called only after the title label creation and when the width of the stackPane that wraps the user node changes.
The title label layout is overwritten only to relocate it properly.

This changes should increase notably the performance of any node wrapped with `Borders`.

Running the test attached to the issue shows now:

```
needs layout = false
needs layout = true
needs layout = false
needs layout = true
```
(that's it, no more infinite layout passes)

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/tools/Borders.java
+++ b/controlsfx/src/main/java/org/controlsfx/tools/Borders.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013, 2015, ControlsFX
+ * Copyright (c) 2013, 2024, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -723,7 +723,7 @@ public final class Borders {
         }
 
         @Override public Node wrap(final Node n) {
-            StackPane pane = new StackPane() {
+            return new StackPane() {
                 Label titleLabel;
                 
                 {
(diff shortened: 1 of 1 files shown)


abhinayagarwal:
I can reproduce this with the provided sample in issue #1538. I can further confirm that this PR fixes the issue. Overall, it looks like a good enhancement. Is there a particular reason why `titleLabel` is currently only relocated and not resized in `layoutChildren` ?

jperedadnr:
I overwrote `computePrefWidth` for that, and that should do it.

Adding `titleLabel.resize` to the layout pass brings back the infinite layout passes.

For the test case for instance, the label width (without padding) is 26.5, while pref width value (without padding) is: 26.4609375. 

Without this change, there is a constant fight between pref and width:

- StrokeBorders layoutChildren sets the label's width to its pref width,  and Region::widthChanged calls requestParentLayout. During the layout, the width gets snapped, and there is this log (adding a log to `Region::widthChanged`) repeated all over again (every pulse):

```
...
needs layout = false
value = 26.5 vs _width = 26.4609375
needs layout = true
value = 26.4609375 vs _width = 26.5
...
```

Of course, and option might be adding snap methods to the layout pass, but I guess `computePrefWidth` already takes care of that.

abhinayagarwal (approved):

Issue #1550, opened 2024-06-10
CheckComboBox title shows "null" when checks are unselected in a certain order


boomah:
![image](https://github.com/controlsfx/controlsfx/assets/366542/f953e543-2d4f-4b1f-96d6-878a3009b637)

Create a CheckComboBox with strings 0 - 9. Select 2, select 4 and then deselect 2. null with be displayed. This is happening in master because of this change:

https://github.com/controlsfx/controlsfx/pull/1541/files

but I don't think that change is incorrect, it just exercises some other code which is incorrect. I believe this is the buggy code:
https://github.com/controlsfx/controlsfx/blob/master/controlsfx/src/main/java/org/controlsfx/control/CheckBitSetModelBase.java#L316
The short circuits in that code don't seem to be working properly. Can be reproduced by adding this test to CheckBitSetModelBaseTest:

`    @Test
    public void testCheckedItemsContainsCorrectValuesAfterClearCheck() {
        model.clearChecks();
        assertTrue(model.isEmpty());

        model.check(2);
        model.check(4);
        assertTrue(model.isChecked(2));
        assertTrue(model.isChecked(4));
        model.clearCheck(2);
        assertFalse(model.isChecked(2));
        assertTrue(model.isChecked(4));

        ObservableList<String> checkedItems = model.getCheckedItems();
        assertEquals(1, checkedItems.size());
        assertTrue(checkedItems.contains(ROW_5_VALUE));
    }`

Workaround is to comment out the short circuits but presumably this has performance implications.

Rizen59:
Hi boomah,

The problem is probably the same with #1549

boomah:
> Hi boomah,
> 
> The problem is probably the same with #1549

yes, expect it is the same issue

stefanodifrancesco:
Seems related to https://github.com/controlsfx/controlsfx/issues/1531 too

credmond:
This control is very, very buggy, which has become apparent after using it more. Trying changing the items afterwards (clear() and add some new). Instant chaos.

danielpeintner:
resolved by https://github.com/controlsfx/controlsfx/pull/1551

Issue #1531, opened 2023-11-20
CheckComboBox: IndexOutOfBoundsException on clearChecks()


dsubelman:
A bug was introduced in v11.2.0 where an `IndexOutOfBoundsException` is thrown when calling `comboBox.getCheckModel().clearChecks()` after an item is removed.

Here's an example that reproduces the issue:

```
import javafx.application.Application;
import javafx.scene.Scene;
import javafx.scene.layout.StackPane;
import javafx.stage.Stage;

import org.controlsfx.control.CheckComboBox;

public class Test extends Application {
    
    @Override
    public void start(Stage stage) throws Exception {
            
        CheckComboBox<String> cb = new CheckComboBox<>();
        cb.getItems().setAll("James", "Emma", "William");
        
        cb.getCheckModel().checkAll();
        cb.getCheckModel().clearCheck(0);
        cb.getItems().remove(0);
        cb.getCheckModel().clearCheck(0);
        cb.getItems().remove(0);
        cb.getCheckModel().clearChecks();
        cb.getItems().clear();
        
        StackPane root = new StackPane(cb);
        
        Scene scene = new Scene(root);
        stage.setScene(scene);
        stage.show();
        
    }
    
    public static void main(String[] args) {
        launch();
    }
}
```

Stack trace:

```
Exception in thread "JavaFX Application Thread" java.util.NoSuchElementException: java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1
	at java.base/java.util.AbstractList$Itr.next(AbstractList.java:377)
	at org.controlsfx.control.CheckBitSetModelBase.lambda$new$2(CheckBitSetModelBase.java:116)
	at javafx.base@20.0.2/com.sun.javafx.collections.ListListenerHelper$SingleChange.fireValueChangedEvent(ListListenerHelper.java:162)
	at javafx.base@20.0.2/com.sun.javafx.collections.ListListenerHelper.fireValueChangedEvent(ListListenerHelper.java:71)
	at javafx.base@20.0.2/javafx.collections.ObservableListBase.fireChange(ObservableListBase.java:238)
	at impl.org.controlsfx.collections.ReadOnlyUnbackedObservableList.callObservers(ReadOnlyUnbackedObservableList.java:44)
	at org.controlsfx.control.CheckBitSetModelBase.lambda$new$1(CheckBitSetModelBase.java:99)
[...]

PR #1551, opened 2024-06-24
update for CheckBitSetModelBase logic

gdicristofaro:
Good afternoon to you,

I encountered an issue when I deselected checkboxes for the `CheckComboBox` in a certain order similar to what was discussed in https://github.com/controlsfx/controlsfx/issues/1550.  Looking through the code, I believe this may fix the issue that I was seeing.  Thanks for the controls you make available here!

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/CheckBitSetModelBase.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/CheckBitSetModelBase.java
@@ -319,13 +319,13 @@ abstract class CheckBitSetModelBase<T> implements IndexedCheckModel<T> {
                 return -1;
             }
 
-            if (index == (lastGetIndex + 1) && lastGetValue < itemCount) {
+            if (index == (lastGetIndex + 1) && index < itemCount) {
                 // we're iterating forward in order, short circuit for
                 // performance reasons (RT-39776)
                 lastGetIndex++;
                 lastGetValue = bitset.nextSetBit(lastGetValue + 1);
                 return lastGetValue;
-            } else if (index == (lastGetIndex - 1) && lastGetValue > 0) {
+            } else if (index == (lastGetIndex - 1) && index > 0) {
                 // we're iterating backward in order, short circuit for
                 // performance reasons (RT-39776)
                 lastGetIndex--;


gdicristofaro:
Ok.  I submitted the CLA.

Siedlerchr (approved):

controlsfx/src/main/java/org/controlsfx/control/CheckBitSetModelBase.java
```diff
@@ -319,13 +319,13 @@ public BitSetReadOnlyUnbackedObservableList(BitSet bitset) {
                 return -1;
             }
 
-            if (index == (lastGetIndex + 1) && lastGetValue < itemCount) {
+            if (index == (lastGetIndex + 1) && index < itemCount) {
```
Rizen59, line 322: index is already tested line 318. The method return if index is superior or equals to itemCount.
So at this step the condition will always be true. Perhaps it could be removed?

danielpeintner:
@abhinayagarwal do you think we can merge this PR (and maybe some others as well) and plan for a release? Thanks!

danielpeintner:
@abhinayagarwal @eugener @Siedlerchr and others
I don't wanna push for anything since I know myself it is sometimes hard to keep up with open source work. Anyhow, I would like to ask whether you still plan to proceed and accept PRs for ControlsFX and plan to publish new versions? Thanks!

abhinayagarwal:
Apologies for the long delay. The change itself looks okay to me.

Can we add a test for this change?

gdicristofaro:
> Can we add a test for this change?

Hi @abhinayagarwal , I just added a few tests.  Let me know if there is anything else.  Thank you for your time in maintaining this project.

danielpeintner:
> Hi @abhinayagarwal , I just added a few tests. Let me know if there is anything else.

@gdicristofaro I wonder whether you checked the example mentioned [here](https://github.com/controlsfx/controlsfx/issues/1531#issuecomment-1929277168)
If it does fix that problem also, we could add this as test case also. Thanks!

gdicristofaro:
> @gdicristofaro I wonder whether you checked the example mentioned [here]...

Sure.  Test is included.

danielpeintner (approved):
LGTM

abhinayagarwal (approved):

PR #1567, opened 2024-09-07
Add task finish button

Siedlerchr:


koppor:
Current shortcoming: The property of `retainTasks` cannot be changed after initialization.

koppor:
Better than nothing I would say. - so +1 from my side

PR #1581, opened 2025-03-06
Add font awesome 6 otf and enum

JurgenCruz:
Added .otf file for fontawesome version 6 and Enum class with the escape sequences for the characters.


JurgenCruz:
Signed the CLA

danielpeintner:
I have a general comment/question.

This PR adds FontAwesome v6 while keeping v4.

According to https://docs.fontawesome.com/web/setup/upgrade/upgrade-from-v4#how-to-upgrade-from-version-4-to-6 it should be possible to replace v4 all together. Maybe I am wrong?

The guide is made for the Web, so it may also not be applicable to JavaFX. Anyhow, if it is possible to replace the old code, I think it would be even better.

JurgenCruz:
Oh, I thought we wanted to keep both as to not force users that don't want to change icons. But I guess it makes sense.

As for the migration, not sure if it will be that straight forward. In the JavaScript world, they used a file named "v4-shims.js" which fixed the backwards compatibility after the icon renaming. The way it worked was having the V6 with the new names, and the V4-shim that mapped the old names to the new ones.

In our case, what do we want to do? do our own shim by keeping the new class and edit the old class to map to the new font with the renames? If that is the case, I would have to manually do it. Plus, at that point if we are needed both classes, might as well leave it and just let the users that want to migrate use the new class and choose any new icons if they so desire. 

Alternatively we could delete the old class and rename the new class as the old. I am not sure if all the old names are in the new version, as that would be the only breaking API if a name was removed. If a user does not like an icon because of the fontawesome renaming, they can just choose a new icon.

I'll let a maintainer choose which of the 3 options we want to move forward with. I personally don't want to make a shim because it would be more extra work for me :stuck_out_tongue:, but I'll do it if that is the decision.

eugener:
In my view, this will be part of the next release. While it may introduce slight incompatibilities, it will also offer enhanced functionality—specifically, more icons in our case. As long as the upgrade path from v4 to v6 is well-documented, developers should be able to assess the necessity of upgrading.

Let’s see what others think.

JurgenCruz:
One thing to notice about the shim. In JS it was needed because you couldn't have both versions at the same time or there would be CSS and JS collisions. This meant you had to completely remove V4, then add V6 and add a **temporary** shim for backwards compatibility. Once you finished migrating, you removed the shim.

In our case, both fonts can coexist perfectly together, and developers can migrate at their own pace after updating with no breakage. We can deprecate the V4 class and remove it in a later release after users have had enough time to migrate. Of course, this only works with option 1. Option 3 would make it backwards incompatible immediately and would force users to do the migration work at the same time they upgrade version.

TL;DR: Shim was needed in JS during migration, not needed in Java during migration. Option 2 is overkill IMO.

danielpeintner:
I don't have a strong preference. Either way is fine by me.

JurgenCruz:
So, just checking. Are we good to merge then?

danielpeintner (approved):
LGTM, even though I think a shim would be even better

PR #1582, opened 2025-03-10
Fix NullPointerException in PopOver occurring randomly on Show call

besidev:
This PR addresses an intermittent `NullPointerException` that occurs during the `fade-in` animation of the PopOver control. Fixes issue #1583 that manifests in the on `show` call when the PopOver’s skin is not yet fully initialized, causing `getSkin()` to return `null`.


besidev:
Signed the CLA

abhinayagarwal:
Hi @besidev,

Please file an issue first and then this PR can fix the issue :)

besidev:
Hi @abhinayagarwal,

Thanks for pointing that out :)
I’ll file the GitHub issue as suggested and update the PR to address it accordingly.

besidev:
@abhinayagarwal 

I’ve created the GitHub issue and updated the PR description to reference it. For additional context, the solution provided in this PR has been successfully tested across various platforms. Could you kindly let me know the expected timeframe for the next ControlsFX release? It would be wonderful if we could include this fix in that release. Thank you very much!

Siedlerchr:
Please update the copyright header at the top, replace 2022 with 2025
Fixes #1583

Siedlerchr (approved):

besidev:
@Siedlerchr @abhinayagarwal 
I've updated the copyright header as requested. Let me know if there's anything else you'd like me to change!

abhinayagarwal:
Currently, `show()` does the following:

1. Setting x, y, width, height of the owner
2. Set things to do after the pop over is shown i.e. `setOnShown` event listener
3. Call `super.show()`
4. Start fade animation
5. Add on hiding event listener

In your current change, you have moved everything into the `onShown` event listener. Any particular reason for this?

besidev:
One of the main reasons for moving most of the logic (the fade-in animation, sizing and positioning) into the `WindowEvent.WINDOW_SHOWN` event handler, rather than doing it immediately after calling `super.show()`, is that a `PopupControl` (and by extension, `PopOver`) does not reliably know that the window is fully displayed at the time `show()` is called. Consequently, if the fade-in animation starts before the popup’s `Scene` and `Skin` are fully initialized, it can lead to null references, such as the `NullPointerException` described in issue #1583. Essentially, this is a timing-based issue that can be difficult to reproduce. By waiting until the popup is truly shown on screen, we ensure that all layout passes are complete and node references are valid before triggering the fade-in animation.

Furthermore, `setOnShown(...)` is a public API method that developers might also use, potentially overriding internal logic if both the library and user code rely on `setOnShown(...)`. Therefore, hooking into `WindowEvent.WINDOW_SHOWN` (for example via `addEventHandler(WindowEvent.WINDOW_SHOWN, ...)`) is both safer and more robust for finalizing layout setting (x, y, width, height) and starting the animation.

abhinayagarwal:
The only disadvantage I see here is that the PopOver might be shown once before the animation actually starts to play causing a visual glitch.

besidev:
@abhinayagarwal 
Your observation is correct. Based on the code, it does seem that the PopOver may appear briefly before the animation starts, a behavior that was already present before this PR, since calling `super.show()` before running the animation could produce a similar effect. However, in our testing we haven't been able to reproduce this visual glitch, neither now nor previously. A better solution would be to properly set up the animation of the PopOver content (perhaps in the skin class), but addressing that would require its own dedicated PR. What's your opinion?

abhinayagarwal:
Like I said, I don't have a strong opinion here. Let's move this fix in. If this comes to bite us again, we can refactor it to fix it for good.

abhinayagarwal (approved):

besidev:
@abhinayagarwal @Siedlerchr 
Could you please let me know when the next release is planned? We cannot use snapshot versions.

abhinayagarwal:
v11.2.2 is now available

danielpeintner:
> v11.2.2 is now available

Great, thank you!

You may want to create the release tag etc. in GitHub 

grafik

abhinayagarwal:
> You may want to create the release tag etc. in GitHub

Done! 
We need to automate these as a part of the release workflow :)

PR #1607, opened 2025-11-24
#1606: fixes TableRowExpanderColumn display/cache bug

### LINKED ISSUE #1606: TableRowExpanderColumn no longer works due to *misbehaving* code
On [openjfx-dev@openjdk.org](https://mail.openjdk.org/pipermail/openjfx-dev/2025-November/057759.html) @credmond reported a problem with TableView and TableRowExpanderColumn

@Maran23 [replied](https://mail.openjdk.org/pipermail/openjfx-dev/2025-November/057763.html) that ControlsFX uses code in a *wrong* fashion.

Unfortunately, the response from @Maran23 cannot be easily read. Hence, I paste here his response as "image".

Image

credmond:
Fixes #1606 

Existing behaviour never worked as expected, nodes were removed from expandedNodeCache (expandedNodeCache.remove(getBean())) everytime a row was collapsed.

This meant that ExpandableTableRowSkin's tableRow.itemProperty() listener was was getting null from the cache, and unable to remove the children, to clean-up the row.

This problem went unnoticed until now because a table refresh was re-creating all cells anyway, so the un-removed nodes were getting discarded anyway. Now though, there are new optimisations in TableView which means these leftover nodes are visible as they're never never cleaned up from the rows.

The fix is to clean them up as I believe was originally intended.

### DIFF
--- a/controlsfx/src/main/java/impl/org/controlsfx/skin/ExpandableTableRowSkin.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/skin/ExpandableTableRowSkin.java
@@ -59,7 +59,7 @@ public class ExpandableTableRowSkin<S> extends TableRowSkin<S> {
         this.expander = expander;
         tableRow.itemProperty().addListener((observable, oldValue, newValue) -> {
             if (oldValue != null) {
-                Node expandedNode = this.expander.getExpandedNode(oldValue);
+                Node expandedNode = this.expander.removeExpandedNode(oldValue);
                 if (expandedNode != null) getChildren().remove(expandedNode);
             }
         });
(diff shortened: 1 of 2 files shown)


Maran23 (approved):
This looks correct! Removing the expanded node when the item changed is the correct approach here.

This should also improve the performance a bit since we are not eagerly removing the node from the cache anymore (on every collapse) - just when we need to - that is when the item changed. Therefore, we are also not recreating the same expanded node more often then we need to.

controlsfx/src/main/java/impl/org/controlsfx/skin/ExpandableTableRowSkin.java
```diff
@@ -59,7 +59,7 @@ public ExpandableTableRowSkin(TableRow<S> tableRow, TableRowExpanderColumn<S> ex
         this.expander = expander;
         tableRow.itemProperty().addListener((observable, oldValue, newValue) -> {
             if (oldValue != null) {
-                Node expandedNode = this.expander.getExpandedNode(oldValue);
+                Node expandedNode = this.expander.removeExpandedNode(oldValue);
```
credmond: Without the fix in place, expandedNode may be null and therefore never removed from the row.

PR #1615, opened 2026-03-18
Fix accumulated width in TableRow2Skin Fixes #1576

### LINKED ISSUE #1576: TableView2 horizontal ScrollBar max value
The max value for the horizontal ScrollBar is being set lower than it should be.  This causes a user to be unable to scroll all the way to the right of a large table.  Appears to be a rounding error that accumulates, as it scales with the number of columns, and is most evident in a table with a large number of columns. Somewhere around 100 columns (depending on width), entire columns are unable to be viewed.  This behavior is present using a TableView2, but not using a TableView, all other things equal.

In this image you can see that a whole column is not shown when using TableView2.  This table has 100 columns

Image

JimmyBoomBots3000:
### Root cause
In`layoutChildren`, each column's width is computed as `snapSizeX(column.getWidth())` (line 185), which rounds up to pixel boundaries. But `computePrefWidth` was summing the raw unsnapped widths. With ~100 columns, this accumulated rounding difference made `computePrefWidth` return a value smaller than the actual layout width, causing the scrollbar max to be set too low and leaving the rightmost columns unreachable.

### Fix
Apply `snapSizeX()` per column in `computePrefWidth`, so it matches the width calculation used in `layoutChildren`.

### DIFF
--- a/controlsfx/src/main/java/impl/org/controlsfx/tableview2/TableRow2Skin.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/tableview2/TableRow2Skin.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013, 2025, ControlsFX
+ * Copyright (c) 2013, 2026, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


jperedadnr (commented):
Minor: You need to update the copyright year.

The fix looks good and clearly solves the issue, given that, as you said, `layoutChildren` already uses `snapSizeX`.

It would be good to include a small test in `TableView2Test`, given that there is already a tableView with 500 columns. If you set a fractional width for these columns, you should be able to easily compute both the row.prefWidth and the sum of the snapped columns width. With your fix, both should match, but differ wihtout it.

JimmyBoomBots3000:
> Minor: You need to update the copyright year.
> 
> The fix looks good and clearly solves the issue, given that, as you said, `layoutChildren` already uses `snapSizeX`.
> 
> It would be good to include a small test in `TableView2Test`, given that there is already a tableView with 500 columns. If you set a fractional width for these columns, you should be able to easily compute both the row.prefWidth and the sum of the snapped columns width. With your fix, both should match, but differ wihtout it.

Updated copyright year and added a unit test.

Issue #1619, opened 2026-07-28
Support Java 26


atehrani-statrad:
When attempting to build and run this project on Java 26 it fails due to using older version of Gradle and respective plugins

PR #1620, opened 2026-07-28
Updates to build/run on Java 26

atehrani-statrad:
Addresses issue #1619

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,13 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: Setup Java
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v5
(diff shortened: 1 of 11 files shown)


.github/workflows/build.yml
```diff
@@ -12,13 +12,13 @@ jobs:
...
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: Setup Java
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v5
         with:
           distribution: 'temurin'
-          java-version: '17'
+          java-version: '26'
```
Siedlerchr: I think we should stick to 25 here

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         uses: actions/setup-java@v5
         with:
           distribution: 'temurin'
-          java-version: '26'
+          java-version: '25'
```
danielpeintner, line 21: 👍
You might want to update the PR description / subject line

PR #1624, opened 2026-09-09
Fix the check model of CheckListView / CheckComboBox losing track of its checks

### LINKED ISSUE #1549: CheckListView.getCheckModel() doesn't select an item if another with a superior index was previously checked
**Description / Steps to reproduce:** 

Run the following code with controlsfx 11.2.1 and click on the button.
As defined in setOnAction(), items 2, 3, 1 should be checked. But only items 2 and 3 are.
If you click again on the button item 1 will be selected.
There is no problem if items are selected in order 1, 2, 3:

**Environment:**

Tested with JDK 21.0.3 / JavaFX 23-ea+20 / controlsfx 11.2.1
No problem with controlsfx 11.1.2, it seems to be a regression from 11.2.0

```java
public class CheckListViewIssue extends Application {

    @Override
    public void start(Stage primaryStage) {
        VBox vbox = new VBox();

        ObservableList<String> items = FXCollections.observableArrayList(
                "Item 1", "Item 2", "Item 3", "Item 4"
        );

        CheckListView<String> checkListView = new CheckListView<>(items);

        Button button = new Button("Select items in this order : 2, 3, 1. Item 1 will not be selected.");
        button.setOnAction(event -> {
            checkListView.getCheckModel().check(items.get(1));
            checkListView.getCheckModel().check(items.get(2));
            checkListView.getCheckModel().check(items.get(0)); // will not be selected
        });

        vbox.getChildren().add(button);
        vbox.getChildren().add(checkListView);

        primaryStage.setScene(new Scene(vbox));
        primaryStage.show();
    }

    public static void main(String[] args) {
        launch(args);
    }
}
```

### LINKED ISSUE #1597: Inconsistent selection in CheckComboBox
Hello,

While using CheckComboBox, I noticed a visual inconsistency bug in the selection of checkboxes.

### Bug description
When you have four or more contiguous items selected and you toggle the selection of the third from last item, it only deselects and does not reselect. Here are some examples:

Image

Image

After seeing this, I decided to add some prints to the CheckBitSetModelBase class and found something quite strange.

### Code
```
CheckBitSetModelBase(final Map<T, BooleanProperty> itemBooleanMap) {

        ...

        // this code is to handle the situation where a developer is manually
        // toggling the check model, and expecting the UI to update (without
        // this it won't happen!).
        getCheckedItems().addListener((ListChangeListener<T>) c -> {
            while (c.next()) {
                if (c.wasAdded()) {
                    for (T item : c.getAddedSubList()) {
                        System.out.println("[c.wasAdded()] item = " + item + ", c.getAddedSubList() = " + c.getAddedSubList()); # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
                        updateBooleanProperty(item, true);
                    }
                } 
                
                if (c.wasRemoved()) {
                    for (T item : c.getRemoved()) {
                        updateBooleanProperty(item, false);
                    }
                }
            }
        });
    }

    private void updateBooleanProperty(T item, boolean value) {
        BooleanProperty p = getItemBooleanProperty(item);
        if (p != null) {
            boolean before = p.get(); # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
            listenerFlag.set(true);
            p.set(value);
            listenerFlag.set(false);
            System.out.println("item = " + item + ", before = " + before + ", after = " + p.get()); # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
        }
    }
```

### Output

Selection of the first four items (everything normal)
```
[c.wasAdded()] item = Item 0, c.getAddedSubList() = [Item 0]
item = Item 0, before = false, after = true
============================================
Change: { [Item 0] addition at 0,  }
Added sublist [Item 0]
Removed sublist []
List [Item 0]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
[c.wasAdded()] item = Item 1, c.getAddedSubList() = [Item 1]
item = Item 1, before = false, after = true
============================================
Change: { [Item 1] addition at 1,  }
Added sublist [Item 1]
Removed sublist []
List [Item 0, Item 1]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
[c.wasAdded()] item = Item 2, c.getAddedSubList() = [Item 2]
item = Item 2, before = false, after = true
============================================
Change: { [Item 2] addition at 2,  }
Added sublist [Item 2]
Removed sublist []
List [Item 0, Item 1, Item 2]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
[c.wasAdded()] item = Item 3, c.getAddedSubList() = [Item 3]
item = Item 3, before = false, after = true
============================================
Change: { [Item 3] addition at 3,  }
Added sublist [Item 3]
Removed sublist []
List [Item 0, Item 1, Item 2, Item 3]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
```

Alternating the third from last selected item, i.e., deselecting it and selecting it again (bug)
```
item = Item 1, before = true, after = false
============================================
Change: { [Item 1] removal at 1,  }
Added sublist []
Removed sublist [Item 1]
List [Item 0, Item 2, Item 3]
Added false, Permutated false, Removed true, Replaced false, Updated false
============================================
[c.wasAdded()] item = Item 2, c.getAddedSubList() = [Item 1]
item = Item 2, before = true, after = true
============================================
Change: { [Item 1] addition at 1,  }
Added sublist [Item 1]
Removed sublist []
List [Item 0, Item 1, Item 2, Item 3]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
```
There is definitely a problem here, because when you deselect the item, it is set to False, but when you try to reselect it, the next item is set to True?

But that's not all. If you add one more print, specifically this one:

```
// this code is to handle the situation where a developer is manually
// toggling the check model, and expecting the UI to update (without
// this it won't happen!).
getCheckedItems().addListener((ListChangeListener<T>) c -> {
    while (c.next()) {
        System.out.println("[c.next()] c = " + c); # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
        if (c.wasAdded()) {
            for (T item : c.getAddedSubList()) {
                System.out.println("[c.wasAdded()] item = " + item + ", c.getAddedSubList() = " + c.getAddedSubList());
                updateBooleanProperty(item, true);
            }
        } 
        
        if (c.wasRemoved()) {
            for (T item : c.getRemoved()) {
                updateBooleanProperty(item, false);
            }
        }
    }
```
The bug disappears:
Image

### Output
```
[c.next()] c = { [Item 0] addition at 0,  }
[c.wasAdded()] item = Item 0, c.getAddedSubList() = [Item 0]
item = Item 0, before = false, after = true
============================================
Change: { [Item 0] addition at 0,  }
Added sublist [Item 0]
Removed sublist []
List [Item 0]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
[c.next()] c = { [Item 1] addition at 1,  }
[c.wasAdded()] item = Item 1, c.getAddedSubList() = [Item 1]
item = Item 1, before = false, after = true
============================================
Change: { [Item 1] addition at 1,  }
Added sublist [Item 1]
Removed sublist []
List [Item 0, Item 1]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
[c.next()] c = { [Item 2] addition at 2,  }
[c.wasAdded()] item = Item 2, c.getAddedSubList() = [Item 2]
item = Item 2, before = false, after = true
============================================
Change: { [Item 2] addition at 2,  }
Added sublist [Item 2]
Removed sublist []
List [Item 0, Item 1, Item 2]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
[c.next()] c = { [Item 3] addition at 3,  }
[c.wasAdded()] item = Item 3, c.getAddedSubList() = [Item 3]
item = Item 3, before = false, after = true
============================================
Change: { [Item 3] addition at 3,  }
Added sublist [Item 3]
Removed sublist []
List [Item 0, Item 1, Item 2, Item 3]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
[c.next()] c = { [Item 1] removal at 1,  }
item = Item 1, before = true, after = false
============================================
Change: { [Item 1] removal at 1,  }
Added sublist []
Removed sublist [Item 1]
List [Item 0, Item 2, Item 3]
Added false, Permutated false, Removed true, Replaced false, Updated false
============================================
[c.next()] c = { [Item 2] addition at 1,  }
[c.wasAdded()] item = Item 1, c.getAddedSubList() = [Item 1]
item = Item 1, before = false, after = true
============================================
Change: { [Item 1] addition at 1,  }
Added sublist [Item 1]
Removed sublist []
List [Item 0, Item 1, Item 2, Item 3]
Added true, Permutated false, Removed false, Replaced false, Updated false
============================================
```

If I find an explanation for this kind of black magic, I'll make a PR, but for now, I don't think a print is the solution. I'm posting this issue in case anyone has any idea where the problem lies.

Rizen59:
## What this fixes

  `CheckListView`, `CheckComboBox` and `CheckTreeView` share `CheckBitSetModelBase`, which holds its
  checks as bits of a `BitSet`. Several defects let that model drift away from what the UI shows:
  `getCheckedIndices()` / `getCheckedItems()` report entries which do not match the ticked boxes, a box
  renders ticked on the wrong row, a check silently disappears, or a listener dies on an
  `IndexOutOfBoundsException`.

  Fixes #1549
  Fixes #1597

  ## The defects

  **1. A stale iteration cache** (the cause of #1549 and #1597) — `BitSetReadOnlyUnbackedObservableList.get(int)`
  caches the last resolved `(position, value)` pair to keep a sequential walk cheap (RT-39776). That pair
  only stays valid as long as the `BitSet` is left untouched, but `check()`, `checkIndices()`,
  `clearCheck()` and `clearChecks()` mutated the `BitSet` directly, behind the list's back. The following
  lookups resumed from an invalid pair and answered a wrong index — visible when a check closes a gap in
  an already checked range, or when the same index is checked twice.

  The list now owns the `BitSet`: no other handle on it remains, the model reads and mutates it through
  the list, and every mutator drops the cached pair. The invalidation can no longer be bypassed.

  **2. The cache left in an exhausted state** — `get(int)` answers `-1` for a position beyond the checked
  count, and used to leave the cached pair pointing at that exhausted state. The forward and backward
  short circuits cannot tell it apart from a resolved one, so the next lookup was computed from it and
  returned a wrong index: reading one index too far corrupted the reads which followed. `get(int)` now
  caches a pair only when it resolved one, and resumes its forward walk from that pair instead of
  counting from the first bit again at every range boundary.

  **3. `checkIndices(int...)` reported its arguments rather than its effect** — it set every given bit and
  then reported the whole argument array as added, without ever looking at it. Out-of-range indices were
  let through where `check(int)` rejects them, an already checked index was reported as added a second
  time, and duplicate or unsorted arguments broke the range change, which walks the reported indices
  assuming they are distinct and ascending. It now reduces the arguments to the indices it actually adds,
  and reports them as one range per run of consecutive indices — so the `splitChanges` parameter and the
  branch reporting a single range spanning unrelated positions are gone.

  **4. No-op calls reported a change** — `check(int)` on an already checked index, `clearCheck(int)` on an
  unchecked one and `clearChecks()` on an empty model each fired a change describing nothing
  (`clearCheck()` looked up a position for an index the list never held). They return early now.

  **5. `isChecked(int)` threw on an index outside the list** — it reached straight into the `BitSet`, so it
  threw `IndexOutOfBoundsException` on the `-1` that `getItemIndex(T)` answers for an item the list does
  not hold; both `isChecked(T)` and `toggleCheckState(T)` hand that `-1` straight over. It answers `false`
  outside the list now, the way `check(int)` and `clearCheck(int)` already ignore such an index.

  **6. Checks did not follow their items when the item list changed** — `updateMap()` rebuilds the
  `BooleanProperty` of every item but left the `BitSet` untouched, so a check stayed on a *position*
  rather than on the item it was made on. Removing a checked item left its bit behind, out of the bounds
  of the list or picked up by whichever item landed on that index — which then showed up checked although
  it never was; `clearChecks()` then mapped that stale index back to an item and threw
  `IndexOutOfBoundsException` inside a listener, where JavaFX only hands the failure over to the uncaught
  exception handler of the thread. And because the rebuild replaces the map rather than the checked
  indices, no listener fired to close the gap: the box of an item the model still reported as checked
  rendered unchecked.

  `updateMap()` now reads the checked items from the `BooleanProperty` of each item — an index is stale as
  soon as the list of items changes, while a property stays with the item it was created for — and
  rebuilds both the checked indices and the properties from them. An item the changed list still holds
  keeps its check, wherever it landed; a check made on an item which is gone dies with it.

  **7. `checkAll()` fired a change per item** — it hands every index over in a single call now, reported as
  one change.

  **8. The added-range change overrode `getAddedSize()`** — the override answered `to - from` without the
  state check its `getFrom()` and `getTo()` make, so an inspection made before `next()` read `0` where
  every other one throws `IllegalStateException`. The base implementation computes the same size through
  those two accessors, so the override is dropped.

  ## Behaviour visible to callers

  - No change event is fired for a call which changes nothing.
  - `checkIndices(int...)` ignores out-of-range indices instead of setting them (`BitSet.set(-1)` threw),
    and reports only the indices it adds.
  - `isChecked(int)` answers `false` outside the list instead of throwing.
  - Checks follow their items across a change of the item list, instead of staying on the positions they
    were made at.

  ## Tests

  `./gradlew :controlsfx:test` — the whole controlsfx suite passes (85 tests, no failure).

  - `CheckBitSetModelBaseTest`: 14 → 33 tests, covering the check orders of #1549 and #1597, the reads
    beyond the checked count, the invalid and duplicate indices, the no-op calls and the item list
    changes.
  - `CheckListViewTest`: new, 9 tests driving a real `CheckListView` and asserting on the
    `BooleanProperty` the `CheckBox` of each row is bound to — the same scenarios, seen from the control.

  The commits are meant to be read in order: each one states the defect it repairs and the reasoning
  behind the fix

### DIFF
--- a/controlsfx/src/main/java/org/controlsfx/control/CheckComboBox.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/CheckComboBox.java
@@ -148,7 +148,10 @@ public class CheckComboBox<T> extends ControlsFXControl {
         this.items = items == null ? FXCollections.observableArrayList() : items;
         checkModelProperty().addListener((o, oldModel, newModel) -> {
             if (oldModel instanceof CheckBitSetModelBase) {
-                ((CheckBitSetModelBase<?>) oldModel).dispose();
+                ((CheckBitSetModelBase<?>) oldModel).detach();
+            }
+            if (newModel instanceof CheckBitSetModelBase) {
+                ((CheckBitSetModelBase<?>) newModel).attach();
             }
         });
         setCheckModel(new CheckComboBoxBitSetCheckModel<>(this.items, itemBooleanMap));
@@ -385,15 +388,22 @@ public class CheckComboBox<T> extends ControlsFXControl {
             super(itemBooleanMap);
             
             this.items = items;
-            this.items.addListener(itemsListener);
-            
-            updateMap();
+
+            attach();
         }
-        
+
         @Override
-        void dispose() {
+        void attach() {
+            // removed first: this model is attached again whenever it is set on its control
             items.removeListener(itemsListener);
-            super.dispose();
+            items.addListener(itemsListener);
+            super.attach();
+        }
+
+        @Override
+        void detach() {
+            items.removeListener(itemsListener);
+            super.detach();
         }
         
         
--- a/controlsfx/src/main/java/org/controlsfx/control/CheckComboBox.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/CheckComboBox.java
@@ -146,6 +146,11 @@ public class CheckComboBox<T> extends ControlsFXControl {
         
         this.itemBooleanMap = new HashMap<>(initialSize);
         this.items = items == null ? FXCollections.observableArrayList() : items;
+        checkModelProperty().addListener((o, oldModel, newModel) -> {
+            if (oldModel instanceof CheckBitSetModelBase) {
+                ((CheckBitSetModelBase<?>) oldModel).dispose();
+            }
+        });
         setCheckModel(new CheckComboBoxBitSetCheckModel<>(this.items, itemBooleanMap));
     }
 
@@ -366,6 +371,7 @@ public class CheckComboBox<T> extends ControlsFXControl {
          **********************************************************************/
         
         private final ObservableList<T> items;
+        private final ListChangeListener<T> itemsListener = c -> updateMap();
         
         
         
@@ -379,9 +385,15 @@ public class CheckComboBox<T> extends ControlsFXControl {
             super(itemBooleanMap);
             
             this.items = items;
-            this.items.addListener((ListChangeListener<T>) c -> updateMap());
+            this.items.addListener(itemsListener);
             
             updateMap();
+        }
+        
+        @Override
+        void dispose() {
+            items.removeListener(itemsListener);
+            super.dispose();
         }
         
         
--- a/controlsfx/src/main/java/impl/org/controlsfx/collections/ChangeHelper.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/collections/ChangeHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, ControlsFX
+ * Copyright (c) 2018, 2026, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -31,12 +31,17 @@ import java.util.List;
 
 class ChangeHelper {
     
-    static String addRemoveChangeToString(int from, int to, List<?> list, List<?> removed) {
+    /**
+     * The text of the given add/remove change, built from the elements it added and the ones it
+     * removed. The added elements are taken as given rather than read back from the list, which a
+     * change reported from a snapshot is free to no longer stand for.
+     */
+    static String addRemoveChangeToString(int from, int to, List<?> added, List<?> removed) {
         
         StringBuilder b = new StringBuilder();
 
         if (removed.isEmpty()) {
-            b.append(list.subList(from, to));
+            b.append(added);
             b.append(" addition at ").append(from);
         } else {
             b.append(removed);
@@ -44,7 +49,7 @@ class ChangeHelper {
                 b.append(" removal at ").append(from);
             } else {
                 b.append(" replaced by ");
-                b.append(list.subList(from, to));
+                b.append(added);
                 b.append(" at ").append(from);
             }
         }
--- a/controlsfx/src/main/java/impl/org/controlsfx/collections/MappingChange.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/collections/MappingChange.java
@@ -138,7 +138,7 @@ public final class MappingChange<E, F> extends Change<F> {
             } else if (wasUpdated()) {
                 b.append(ChangeHelper.updateChangeToString(getFrom(), getTo()));
             } else {
-                b.append(ChangeHelper.addRemoveChangeToString(getFrom(), getTo(), getList(), getRemoved()));
+                b.append(ChangeHelper.addRemoveChangeToString(getFrom(), getTo(), getAddedSubList(), getRemoved()));
             }
             if (pos != size) {
                 b.append(", ");
--- a/controlsfx/src/main/java/impl/org/controlsfx/collections/NonIterableChange.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/collections/NonIterableChange.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, ControlsFX
+ * Copyright (c) 2018, 2026, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -92,7 +92,7 @@ public abstract class NonIterableChange<E> extends Change<E> {
         } else if (wasUpdated()) {
             string = ChangeHelper.updateChangeToString(from, to);
         } else {
-            string = ChangeHelper.addRemoveChangeToString(from, to, getList(), getRemoved());
+            string = ChangeHelper.addRemoveChangeToString(from, to, getAddedSubList(), getRemoved());
         }
         invalid = tempInvalid;
         return "{ " + string + " }";
--- a/controlsfx/src/main/java/impl/org/controlsfx/skin/CheckComboBoxSkin.java
+++ b/controlsfx/src/main/java/impl/org/controlsfx/skin/CheckComboBoxSkin.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013, 2022, ControlsFX
+ * Copyright (c) 2013, 2026, ControlsFX
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -26,7 +26,6 @@
  */
 package impl.org.controlsfx.skin;
 
-import impl.org.controlsfx.collections.ReadOnlyUnbackedObservableList;
 import javafx.beans.binding.Bindings;
 import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
@@ -44,6 +43,9 @@ import javafx.scene.input.KeyCombination;
 import org.controlsfx.control.CheckComboBox;
 
 import org.controlsfx.control.IndexedCheckModel;
+
+import java.util.Collections;
+import java.util.List;
 
 import static org.controlsfx.control.CheckComboBox.COMBO_BOX_ROWS_TO_MEASURE_WIDTH_KEY;
 
@@ -70,8 +72,8 @@ public class CheckComboBoxSkin<T> extends SkinBase<CheckComboBox<T>> {
     // data
     private final CheckComboBox<T> control;
     private final ObservableList<T> items;
-    private final ReadOnlyUnbackedObservableList<Integer> selectedIndices;
-    private final ReadOnlyUnbackedObservableList<T> selectedItems;
+    // the listener through which this skin follows the checks of the model the control has
+    private final ListChangeListener<Integer> checkedIndicesListener;
     
     
     /**************************************************************************
@@ -86,9 +88,6 @@ public class CheckComboBoxSkin<T> extends SkinBase<CheckComboBox<T>> {
         
         this.control = control;
         this.items = control.getItems();
-        
-        selectedIndices = (ReadOnlyUnbackedObservableList<Integer>) control.getCheckModel().getCheckedIndices();
-        selectedItems = (ReadOnlyUnbackedObservableList<T>) control.getCheckModel().getCheckedItems();
         
         comboBox = new ComboBox<T>(items) {
             @Override
@@ -135,7 +134,15 @@ public class CheckComboBoxSkin<T> extends SkinBase<CheckComboBox<T>> {
 
         // The zero is a dummy value - it just has to be legally within the bounds of the
         // item count for the CheckComboBox items list.
-        selectedIndices.addListener((ListChangeListener<Integer>) c -> buttonCell.updateIndex(0));
+        checkedIndicesListener = c -> buttonCell.updateIndex(0);
+
+        // the button renders the checks of the model the control has, which an application is
+        // free to replace through CheckComboBox.setCheckModel(IndexedCheckModel)
+        followCheckModel(null, control.getCheckModel());
+        control.checkModelProperty().addListener((o, oldModel, newModel) -> {
+            followCheckModel(oldModel, newModel);
+            buttonCell.updateIndex(0);
+        });
         
         getChildren().add(comboBox);
     }
@@ -204,7 +211,7 @@ public class CheckComboBoxSkin<T> extends SkinBase<CheckComboBox<T>> {
             String vResult = control.getTitle();
             if (control.isShowCheckedCount()) {
                 //...adding also the count of how many are selected, if so configured
-                vResult = String.format("%s (%d/%d)", vResult, selectedItems.size(), items.size());
+                vResult = String.format("%s (%d/%d)", vResult, getCheckedItems().size(), items.size());
             }
             return vResult;
         } else {            
@@ -214,10 +221,27 @@ public class CheckComboBoxSkin<T> extends SkinBase<CheckComboBox<T>> {
         
     }
     
+    /** Follows the checked indices of the model the control is given, and only of that one. */
+    private void followCheckModel(IndexedCheckModel<T> droppedModel, IndexedCheckModel<T> newModel) {
+        if (droppedModel != null) {
+            droppedModel.getCheckedIndices().removeListener(checkedIndicesListener);
+        }
+        if (newModel != null) {
+            newModel.getCheckedIndices().addListener(checkedIndicesListener);
+        }
+    }
+
+    /** The items checked in the model the control has - what the button is to render. */
+    private List<T> getCheckedItems() {
+        final IndexedCheckModel<T> checkModel = control.getCheckModel();
+        return checkModel == null ? Collections.emptyList() : checkModel.getCheckedItems();
+    }
+
     private String buildString() {
         final StringBuilder sb = new StringBuilder();
-        for (int i = 0, max = selectedItems.size(); i < max; i++) {
-            T item = selectedItems.get(i);
+        final List<T> checkedItems = getCheckedItems();
+        for (int i = 0, max = checkedItems.size(); i < max; i++) {
+            T item = checkedItems.get(i);
             if (control.getConverter() == null) {
                 sb.append(item);
             } else {
--- a/controlsfx/src/main/java/org/controlsfx/control/CheckBitSetModelBase.java
+++ b/controlsfx/src/main/java/org/controlsfx/control/CheckBitSetModelBase.java
@@ -253,8 +253,10 @@ abstract class CheckBitSetModelBase<T> implements IndexedCheckModel<T> {
             setItemChecked(item, true);
         }
 
-        updateFollowedProperties();
+        // reported before the properties are written: a listener of a property this model is
+        // clearing is free to reach back into it, and has to be answered the checks it now holds
         fireChangesAfterItemsChanged();
+        updateFollowedProperties();
     }
 
     /**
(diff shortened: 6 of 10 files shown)


Rizen59:
This PR addresses #1549.

Although this fix was generated by Claude Code, all unit tests (CheckListViewTest and CheckBitSetModelBaseTest) are based on the concrete cases described in #1549.
I also tested #1597 and can confirm that the reported case is fixed.

Rizen59:
I found two new particular cases corresponding to these situations. I'll fix it in couple of minutes : 

```java
public class CheckListViewIssue extends Application {

    @Override
    public void start(Stage primaryStage) {
        VBox vbox = new VBox();

        ObservableList<String> items = FXCollections.observableArrayList(
                "Item 1", "Item 2", "Item 3", "Item 4", "Item 5"
        );

        CheckListView<String> checkListView = new CheckListView<>(items);

        Button button = new Button("Select items 5 and 1 with checkIndices(4, 0). Item 1 will not be selected.");
        button.setOnAction(event -> {
            // the indices are not in ascending order: an IndexOutOfBoundsException is logged and
            // item 1 is checked in the model, rendered unchecked
            checkListView.getCheckModel().checkIndices(4, 0);
        });

        vbox.getChildren().add(button);
        vbox.getChildren().add(checkListView);

        primaryStage.setScene(new Scene(vbox));
        primaryStage.show();
    }

    public static void main(String[] args) {
        launch(args);
    }
}
```

```java
public class CheckListViewIssue extends Application {

    @Override
    public void start(Stage primaryStage) {
        VBox vbox = new VBox();

        ObservableList<String> items = FXCollections.observableArrayList(
                "Item 1", "Item 2", "Item 3", "Item 4", "Item 5"
        );

        CheckListView<String> checkListView = new CheckListView<>(items);

        Button button = new Button("Select item 5 with checkIndices(4, 4). Item 5 will not be selected.");
        button.setOnAction(event -> {
            // the same index is given twice: an IndexOutOfBoundsException is logged and item 5 is
            // checked in the model, rendered unchecked
            checkListView.getCheckModel().checkIndices(4, 4);
        });

        vbox.getChildren().add(button);
        vbox.getChildren().add(checkListView);

        primaryStage.setScene(new Scene(vbox));
        primaryStage.show();
    }

    public static void main(String[] args) {
        launch(args);
    }
}
```

Rizen59:
I found more issues in this class, this time involving the internal checkedIndicesList, which was getting out of sync. The effects can be seen in the following mini-apps. As with the other cases, I turned them into unit tests in CheckBitSetModelBaseTest and validated them manually before implementing the fix. The next commit fixes this.

```java
public class CheckListViewIssue extends Application {

    @Override
    public void start(Stage primaryStage) {
        VBox vbox = new VBox();

        ObservableList<String> items = FXCollections.observableArrayList(
                "Item 1", "Item 2", "Item 3", "Item 4", "Item 5"
        );

        CheckListView<String> checkListView = new CheckListView<>(items);

        ListView<String> boundItems = new ListView<>();
        Bindings.bindContent(boundItems.getItems(), checkListView.getCheckModel().getCheckedItems());

        Button button = new Button("Select item 2, then select item 2 again. It is listed twice in the bound list below.");
        button.setOnAction(event -> {
            checkListView.getCheckModel().check(items.get(1));
            checkListView.getCheckModel().check(items.get(1)); // already checked, no-op on the model, still reported as added
        });

        vbox.getChildren().add(button);
        vbox.getChildren().add(checkListView);
        vbox.getChildren().add(new Label("List bound to getCheckedItems() :"));
        vbox.getChildren().add(boundItems);

        primaryStage.setScene(new Scene(vbox));
        primaryStage.show();
    }

    public static void main(String[] args) {
        launch(args);
    }
}
```

```java
public class CheckListViewIssue extends Application {

    @Override
    public void start(Stage primaryStage) {
        VBox vbox = new VBox();

        ObservableList<String> items = FXCollections.observableArrayList(
                "Item 1", "Item 2", "Item 3", "Item 4", "Item 5"
        );

        CheckListView<String> checkListView = new CheckListView<>(items);

        ListView<String> boundItems = new ListView<>();
        Bindings.bindContent(boundItems.getItems(), checkListView.getCheckModel().getCheckedItems());

        Label exception = new Label("No exception yet.");
        // JavaFX hands the exceptions thrown by a list listener - the content binding above
        // included - over to the uncaught exception handler of the thread, where they go unnoticed
        Thread.currentThread().setUncaughtExceptionHandler((thread, throwable) -> exception.setText(throwable.toString()));

        Button button = new Button("Select items 1 and 3, then unselect item 2 which is not selected. The bound list fails with an IndexOutOfBoundsException.");
        button.setOnAction(event -> {
            checkListView.getCheckModel().check(items.get(0));
            checkListView.getCheckModel().check(items.get(2));
            checkListView.getCheckModel().clearCheck(items.get(1)); // not checked, no-op on the model, still reported as removed from position -1
        });

        vbox.getChildren().add(button);
        vbox.getChildren().add(checkListView);
        vbox.getChildren().add(new Label("List bound to getCheckedItems() :"));
        vbox.getChildren().add(boundItems);
        vbox.getChildren().add(exception);

        primaryStage.setScene(new Scene(vbox));
        primaryStage.show();
    }

    public static void main(String[] args) {
        launch(args);
    }
}
```

```java
public class CheckListViewIssue extends Application {

    @Override
    public void start(Stage primaryStage) {
        VBox vbox = new VBox();

        ObservableList<String> items = FXCollections.observableArrayList(
                "Item 1", "Item 2", "Item 3", "Item 4", "Item 5"
        );

        CheckListView<String> checkListView = new CheckListView<>(items);

        Label result = new Label("Nothing read yet.");

        Button button = new Button("Select items 1 and 2, then read the checked index at position 3 twice. The second read answers 0.");
        button.setOnAction(event -> {
            checkListView.getCheckModel().check(items.get(0));
            checkListView.getCheckModel().check(items.get(1));

            ObservableList<Integer> checkedIndices = checkListView.getCheckModel().getCheckedIndices();
            // position 3 is beyond the two checked indices, so there is no checked index to answer
            // with, but the read still leaves an iteration state the next one starts from
            int first = checkedIndices.get(3);
            int second = checkedIndices.get(3);

            result.setText("get(3) = " + first + ", then get(3) = " + second + " (expected -1 both times)");
        });

        vbox.getChildren().add(button);
        vbox.getChildren().add(result);
        vbox.getChildren().add(checkListView);

        primaryStage.setScene(new Scene(vbox));
        primaryStage.show();
    }

    public static void main(String[] args) {
        launch(args);
    }
}
```

```java
public class CheckListViewIssue extends Application {

    @Override
    public void start(Stage primaryStage) {
        VBox vbox = new VBox();

        ObservableList<String> items = FXCollections.observableArrayList(
                "Item 1", "Item 2", "Item 3", "Item 4", "Item 5"
        );

        CheckListView<String> checkListView = new CheckListView<>(items);

        Label result = new Label("Nothing read yet.");

        Button button = new Button("Select items 1 and 2, read the checked index at position 3, then the one at position 1. The second read throws.");
        button.setOnAction(event -> {
            checkListView.getCheckModel().check(items.get(0));
            checkListView.getCheckModel().check(items.get(1));

            ObservableList<Integer> checkedIndices = checkListView.getCheckModel().getCheckedIndices();
            // position 3 is beyond the two checked indices: the read answers -1 and leaves an
            // iteration state the read of position 1 then walks back from
            checkedIndices.get(3);

            try {
                result.setText("get(1) = " + checkedIndices.get(1) + " (expected 1)");
            } catch (RuntimeException e) {
                result.setText(e + " (expected 1)");
            }
        });

        vbox.getChildren().add(button);
        vbox.getChildren().add(result);
        vbox.getChildren().add(checkListView);

        primaryStage.setScene(new Scene(vbox));
        primaryStage.show();
    }

    public static void main(String[] args) {
        launch(args);
    }
}
```

danielpeintner:
Thank you for your work.

Question: Do we still have the situation that we cannot properly add JUnit test cases?
Having some reproducer examples is fine, but having an actual unit test catching these issues (also w.r.t. to future changes) is important; otherwise we keep circling with manual testing.

Rizen59:
No, that no longer holds — the PR automates every one of these tests, there is nothing left to run by hand.

The CheckListViewIssue snippets above are simple, concrete reproducers, meant to let anyone see each problem on screen. Every one of them has been transcribed into code, in CheckBitSetModelBaseTest (14 → 33 tests) and CheckListViewTest (new, 9 tests). Checked out on master (c268047d) with only those two test files and without the fix, 25 of the 42 tests fail; with the fix, the whole suite passes (85 tests, no failure).

CheckListViewTest replays the same scenarios on a real shown CheckListView via TestFX; I may drop that class in the end, as it partly duplicates CheckBitSetModelBaseTest, but I would rather keep it until I stop finding anomalies.

I moved the PR back to draft for that reason: I have found further problems in this area and want to push them before a final merge, each with the unit test written, reviewed and confirmed failing before the fix.

danielpeintner:
> No, that no longer holds — the PR automates every one of these tests, there is nothing left to run by hand.
> 
> The CheckListViewIssue snippets above are simple, concrete reproducers, meant to let anyone see each problem on screen. Every one of them has been transcribed into code, in CheckBitSetModelBaseTest (14 → 33 tests) and CheckListViewTest (new, 9 tests). Checked out on master ([c268047](https://github.com/controlsfx/controlsfx/commit/c268047db313fac99d57371d7c17b55be033da58)) with only those two test files and without the fix, 25 of the 42 tests fail; with the fix, the whole suite passes (85 tests, no failure).

I see, thank you. 👍

What wonders me a bit is that the CI does (did) not show any issue before your changes.
Maybe someone needs to activate the JUnit test task on the CI so that anyone notices there is a problem with a change.
@abhinayagarwal ?

danielpeintner:
I noticed you pushed a new commit (https://github.com/controlsfx/controlsfx/pull/1624/changes/2ca6992cddcb6a1c5a7951c708e743a0e0d9d635) after I made my comment.
Are those the tests you were referring to, and I shall remove the comment [here](https://github.com/controlsfx/controlsfx/pull/1624#issuecomment-5620034239)

BTW, I think a maintainer needs to approve the workflow run

Rizen59:
No, those aren't the tests I meant: the ones I was referring to were already pushed when I wrote that message. 2ca6992c is only a refactoring commit.

My previous message was probably ambiguous on one point. Saying "Checked out on master (c268047) with only those two test files and without the fix, 25 of  the 42 tests fail" could read as if those tests were already on master. They are not: they come from this branch, I merely dropped them onto master to confirm they fail without the fix. On master as it stands, everything is green. No test covered these bugs, so the CI had nothing to report.

A word about the size of the diff. CheckBitSetModelBase.java has changed a lot, far more than I expected when I started. I went from the concrete cases of #1549 and #1597, and once the first tests were written and those two cases fixed, the automated code review of Claude Code that I use surfaced further anomalies in the same class: checks shifting when an item is added to the list, listeners not reflecting what had actually just been clicked, and others. Each time I was able to reproduce them by hand — hence the reproducers above — and each one ended up as an automated test.

To be clear about the method, here is how I proceed:

-   writing the unit tests, manual review and validation of the problem on a real control;
-   implementing the fix with Claude Code;
-   /simplify, to simplify and optimise the code;
-   /code-review, for the automated review.

If that last step surfaces new problems, I go back to step 1. That loop is what explains the size of the diff, and above all what gives this class a proper test base at last — that is as much the goal here as the two original tickets.

Once things have settled, I will run the whole implementation process again from scratch, to compare it with what was produced along the way. The aim is a final fix which is robust and, where possible, optimised, rather than an accumulation of successive patches. And before that final implementation, I will check that every test is indeed red on master, one by one rather than just "the suite fails", to make sure none of them passes for the wrong reasons and that each one really describes a defect of the current code.

So the PR stays in draft, more commits are coming to finalise it. I'll let you know here as soon as I'm done.

Also thanks for your quick replies and for following this one so closely :blush:

controlsfx/src/main/java/org/controlsfx/control/CheckComboBox.java
```diff
@@ -146,6 +146,14 @@ public CheckComboBox(final ObservableList<T> items) {
         
         this.itemBooleanMap = new HashMap<>(initialSize);
         this.items = items == null ? FXCollections.observableArrayList() : items;
+        checkModelProperty().addListener((o, oldModel, newModel) -> {
+            if (oldModel instanceof CheckBitSetModelBase) {
+                ((CheckBitSetModelBase<?>) oldModel).detach();
+            }
```
danielpeintner, line 152: Minor simplification proposal here and later ... unless we use an old Java version.. not sure

```suggestion
            if (oldModel instanceof CheckBitSetModelBase cbsmb) {
                cbsmb.detach();
            }
```

Rizen59: This Java improvement was introduced in version 16.

This code was generated by Claude Code, but I've planned a check at the end of my workflow to make sure the code stays compatible with Java 11 at minimum (I remember, a long time ago, that ControlsFX went from version 8 to 11 precisely to be compatible with Java 11). Has that changed since?

More commits are on the way — in particular, I'm currently doing a full pass over all the unit tests that were added, to make sure they test a contract issue and not something implementation-specific (which can always change).

danielpeintner: I am not sure either, but I would argue Java 17 is fine. It is used in GitHub CI workflows also
@abhinayagarwal ?

PR #1626, opened 2026-09-16
Add selection batching to reduce change events, including tests

### LINKED ISSUE #1625: TableView2: Shift+Click selection of rows is too slow and freezes the FX thread
Run the attached test app with ControlsFX 11.2.5:

```
import java.util.List;
import java.util.stream.IntStream;

import org.controlsfx.control.tableview2.TableColumn2;
import org.controlsfx.control.tableview2.TableView2;

import javafx.application.Platform;
import javafx.beans.property.SimpleStringProperty;
import javafx.collections.FXCollections;
import javafx.collections.ListChangeListener;
import javafx.geometry.Insets;
import javafx.scene.Scene;
import javafx.scene.control.SelectionMode;
import javafx.scene.control.TableColumn;
import javafx.scene.input.KeyEvent;
import javafx.scene.input.MouseEvent;
import javafx.scene.layout.Priority;
import javafx.scene.layout.VBox;
import javafx.stage.Stage;

public final class TestControlsFXApp  extends Application {
    private static final int ROW_COUNT = 500;
    private static final int COLUMN_COUNT = 100;
    private static final boolean CELL_SELECTION_ENABLED = true;

    private TableView2<RowItem> tableView;
    private long mGestureStartNanos;

    @Override
    public void start(final Stage primaryStage) {
        tableView = new TableView2<>();
        tableView.setRowHeaderVisible(true);
        tableView.setFixedCellSize(24);
        tableView.getSelectionModel().setSelectionMode(SelectionMode.MULTIPLE);
        tableView.getSelectionModel().setCellSelectionEnabled(CELL_SELECTION_ENABLED);

        final List<TableColumn<RowItem, String>> columns = IntStream.range(0, COLUMN_COUNT)
            .mapToObj(this::createColumn)
            .toList();
        tableView.getColumns().addAll(columns);

        tableView.setItems(FXCollections.observableArrayList(
            IntStream.range(0, ROW_COUNT).mapToObj(RowItem::new).toList()));

        tableView.addEventFilter(MouseEvent.MOUSE_PRESSED, e -> mGestureStartNanos = System.nanoTime());
        tableView.addEventFilter(KeyEvent.KEY_PRESSED, e -> mGestureStartNanos = System.nanoTime());
        tableView.getSelectionModel().getSelectedCells().addListener((ListChangeListener<Object>) change ->
            Platform.runLater(this::reportElapsed));

        final VBox root = new VBox(10, tableView);
        root.setPadding(new Insets(10));
        VBox.setVgrow(tableView, Priority.ALWAYS);

        primaryStage.setScene(new Scene(root, 1200, 750));
        primaryStage.show();
    }

    private void reportElapsed() {
        final long elapsedMs = (System.nanoTime() - mGestureStartNanos) / 1_000_000;
        final int selectedCount = tableView.getSelectionModel().getSelectedCells().size();
        System.out.println("Last selection change: " + selectedCount + " cells selected, took " + elapsedMs + " ms ");
    }

    private TableColumn<RowItem, String> createColumn(final int index) {
        final TableColumn2<RowItem, String> column = new TableColumn2<>("Col " + index);
        column.setCellValueFactory(row -> row.getValue().values[index]);
        column.setPrefWidth(90);
        return column;
    }

    private record RowItem(SimpleStringProperty[] values) {
        private RowItem(final int rowIndex) {
            this(new SimpleStringProperty[COLUMN_COUNT]);
            for (int i = 0; i < COLUMN_COUNT; i++) {
                values[i] = new SimpleStringProperty("R" + rowIndex + "C" + i);
            }
        }
    }
}
```
click on row header 1
Shift+click on row header 499, on an M1 Mac, it takes almost two minutes to complete the selection of 49900 cells:
```
Last selection change: 49900 cells selected, took 117881 ms 
```

Then shift+click on row header 500, selecting 100 extra cells takes almost 4 minutes:
```
Last selection change: 50000 cells selected, took 233535 ms 
```

jperedadnr:
Fixes #1625 

Before this PR, selecting cells from the RowHeader (with shift+click) was freezing the FX thread, as it was `O(rows^2 * cols)` (every cell was selected/cleared one by one, firing selection/deselection events that went over and over the list of headers that included duplicated cells).

This PR adds deduplication of headers and batching to selection (adding/removing cells/rows), particularly oriented to cell selection mode, when the number of affected indices is potentially `rows * cols`.

Nine tests have been added (4 of them fail before the PR).


eugener:
Thought I chime in here with few more efficiency ideas ..

**1. One partial row disables the fast path**

`RowHeader.clearRowsSelection()` uses `allMatch(count == columnCount)`. One partially selected row makes the whole check false. The code then falls back to `rowsToDeselect.forEach(sm::clearSelection)`.

That call is very expensive. In `TableView.java:2771`, `clearSelection(row, null)` reads all selected cells. Then it calls `SelectedCellsMap.remove()` per `cell. remove()` does `selectedCells.remove(tp)`, a linear search, and fires one event. Cost is O(cells x columns).

Fix: do not make the path all-or-nothing. Always clear once, then rebuild:
1. `sm.clearSelection()`
2. `sm.selectIndices(first, rest)` for the fully selected rows that stay
3. `sm.select(row, column)` only for the cells of partial rows

Cost becomes 2 events plus the number of partial cells. Partial rows are few in practice.

**2. Row selection mode still clears row by row**

The non-cell branch of `clearRowsSelection()` is also `rowsToDeselect.forEach(sm::clearSelection)`. Each event makes the inner header table run `updateHeaders()` again. This is O(rows²).

Fix: use the same batch. Clear once, then `selectIndices()` for the rows that stay.

**3. Accumulate the sub-changes**

`applyListChange()` calls `clearRowsSelection()` inside the `while (c.next())` loop. Each call reads all selected cells. Collect all removed and added indices first. Then apply one batch.

**4. Small items**

- `isValidIndex()` reads `getItems()` on each index; 
- `selectIndices()` builds the int[] through a stream; 
- `Collectors.toMap` boxes one Integer per cell.

jperedadnr:
Thanks @eugener for your feedback!
I've addressed it:
- changed `isValidIndex()` to `validRows()`, to include in the same call the deduplication, and removed `selectValidIndices().
-  In `applyListChanges()` I collect the removed/added lists before applying the changes
- The major change is in `clearRowsSelection()`: Two paths, one for row selection, one for cell selection.
-- For row selection, it applies one by one row deselection or batching based on the number of rows in the process: for a small number or rows, or when it is low compared to the selected rows one by one is faster and has a small number of change events. Else, batching is faster and leads to two events (clear full selection and select the batch), including restoring of selected index and focus. Added `clearRows()` to handle this.
-- For cell selection, more or less the same, but a little bit more complex, given that each row now can have different scenarios: fully selected, or partially selected with a low or a high number of cells to clear. Added `clearCells()` to handle this. Same as for rows, fast path to remove cells one by one. Else, clear selection, restore full selected cells, and then select individual cells for empty rows, and clear individual cells from selected rows, restoring index and focus.

Have a look!

eugener:
Ran the review on PR :

**1. Focus moves to the wrong cell (regression)**

`RowHeader.java:411`

```java
if (focusedCell != null && cellsToKeepPerRow.containsKey(focusedCell.getRow())) {
    fm.focus(focusedCell.getRow(), focusedCell.getTableColumn());
}
```

`clearCells` calls `m.clearSelection()`, which sets the focus to -1. Then `selectIndices` calls `select(lastIndex)`, which puts the focus on the last kept row. The guard skips the focus restore when the focused row is one of the deselected rows. So the focus stays on the wrong cell.

The user keeps the arrow keys on a row that they did not touch. The focus is independent of the selection, so the guard is not necessary. `clearRows (line 366)` restores the focus unconditionally for row selection mode. Make `clearCells` do the same.

**2. A single row restore is not batched**

`RowHeader.java:269`

```java
if (indices.size() == 1) {
    selectionModel.select(first);
    return;
}
```

In cell selection mode `select(row)` passes a null column. JavaFX then loops over the visible leaf columns and selects each cell one by one (TableView.java:2508-2513). This fires one event per column.

This hits the most common gesture: a plain click on a row header after a range selection.

`selectionModel.selectIndices(first, first)` batches the positions into one `addAll` The duplicate index is removed by the LinkedHashSet.

This also makes `CellSelectionRestorePath.eventCount()` wrong. It returns 0 for this case, but the real cost is one event per column. The safeguard at line 387 compares against a bad estimate.

Not a regression: master is equally slow here.

**Minor**

- Line 387: `rowsToDeselect.size() * columnCount` assumes that each deselected row is fully selected. For partially selected rows the one-by-one cost is lower than this estimate.
- `CellSelectionRestorePath` is an inner class, but it uses no field of `RowHeader`. Give it its own type parameter and make it static.

jperedadnr:
Thanks for the feedback, I've reviewed it and made the needed changes. Have a look again when you have the time.

